Issue #31, opened 2015-03-12
INIReader: Pass parameters as const references


GoogleCodeExporter:
```
Most methods pass their parameters by-value, creating an unneccessary copy. 
Passing the strings as references would be more efficient.

Using c++11 is would also be possible to mark the Get-Functions as const.
```

Original issue reported on code.google.com by `mieisc...@gmail.com` on 19 Aug 2014 at 4:00

Attachments:
- [inih-constref.diff](https://storage.googleapis.com/google-code-attachments/inih/issue-31/comment-0/inih-constref.diff)

PR #42, opened 2015-12-14
C++ code optimization

cosmy1:


benhoyt:
Hi @cosmy1, thanks for the pull request, however, someone suggested this a couple of years ago (see issue #31) and I reiterate what I said there:

> I'm not a C++ guru, but as I understand it modern compilers and implementations 
> of the STL are copy-on-write so this is what happens anyway. Would it be more 
> efficient? You'd have to benchmark this. But I'm kind of hesitant to change 
> this just for the sake of a tiny bit of performance (are a few nanoseconds 
> important when parsing INI files)? Further thoughts or benchmarks welcome.

cosmy1:
Wel in this specific case there is no significant performance (nanoseconds order).
But, since you really don't need these local copies is best to keep the reference rather than a copy.

Look here:
```cpp
string INIReader::MakeKey(string section, string name)
{
    string key = section + "=" + name;
    // Convert to lower case to make section/name lookups case-insensitive
    std::transform(key.begin(), key.end(), key.begin(), ::tolower);
    return key;
}

string INIReader::Get(string section, string name, string default_value)
{
    string key = MakeKey(section, name);
    return _values.count(key) ? _values[key] : default_value;
}

long INIReader::GetInteger(string section, string name, long default_value)
{
    string valstr = Get(section, name, "");
    const char* value = valstr.c_str();
    char* end;
    // This parses "1234" (decimal) and also "0x4D2" (hex)
    long n = strtol(value, &end, 0);
}

```
When you call ```GetInteger(...)``` you are copying locally ```section``` and ```name``` 3 times with no sense (1 in ```GetInteger(...)```, 1 in ```Get(...)```, 1 in ```MakeKey(...)```).

There is no reason to do so. Instead, passing them by constant references just move the data through parameters till them reach ```MakeKey(...)``` which  internally create a new string using those costant references.

Hope this clarify.

benhoyt:
As I mentioned, I'm no C++ guru, but it looks like there's a fair bit of debate over the whys and wherefores of this in modern C++, as to which is actually better or more efficient, [for example](http://stackoverflow.com/questions/10231349/are-the-days-of-passing-const-stdstring-as-a-parameter-over). One of the things I dislike about C++ is that it encourages lots of debate about things like this. :-)

It does look like the const ref approach would be better here, as most of those functions don't need copies. However, given the following, I'd like to keep it as is:

1) The significant debate about which approach is preferred and more efficient
2) It's syntactically simpler as is, and simpler to reason about
3) In any use case I can think of, saving nanoseconds when reading .INI files isn't a concern
4) I see `INIReader` as a simple C++ wrapper, and want to keep the code very simple

Thanks!

cosmy1:
Yeah! I considered just the most optimized and straight solution from a C++ point of view but forgot to keep in mind the simplicity of this library.

PR #44, opened 2015-12-16
Take first of either : or = as the name / value delimiter

jgroffen:
I was looking for a C based ini parser to parse a file that I was already parsing with Python ConfigParser. Your library came in very handy!

My ini file uses colons for name/value delimiting, and I happen to have values that include an equal sign.

I found that an = in the value would override the : as a name/value delimiter. I made a change so that if both : and = appear in the ini entry, the first to occur is treated as the name/value delimiter.

Example:
somename: somevalue=morevalue

Before patch:
The name would parse as 'somename: somevalue'

After patch:
The name would parse as 'somename' and the value would parse as 'somevalue=morevalue'

I tried various combinations to check they work as expected after the patch:
somename=somevalue:morevalue
somename=somevalue=morevalue
somename:somevalue:morevalue

When I say 'work as expected', I mean as ConfigParser (python v2.7) behaves.

Thanks!

### DIFF
--- a/ini.c
+++ b/ini.c
@@ -130,8 +130,11 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
         else if (*start && *start != ';') {
             /* Not a comment, must be a name[=:]value pair */
             end = find_char_or_comment(start, '=');
-            if (*end != '=') {
-                end = find_char_or_comment(start, ':');
+            /* There may be a colon as well, use the first : or = found */
+            if (*(find_char_or_comment(start, ':')) == ':') {
+                if(*end != '=' || end > find_char_or_comment(start, ':')) {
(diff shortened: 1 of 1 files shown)


benhoyt:
Thanks! This looks good. I'm away on vacation for a couple of weeks, and want to take a closer look, but I think this makes sense especially if it's `ConfigParser` behaviour.

benhoyt:
Thanks again for the bug report. I've fixed this in https://github.com/benhoyt/inih/commit/e6735fff642aa5a4cee40939c0ae4b704b28b0a8 and added tests (though it's a slightly different fix than yours, hopefully a bit simpler).

jgroffen:
You're welcome, I'll update to your fix in a few days when I'm back from holidays.

Issue #40, opened 2015-10-18
Unix-style (#) comments in lines seems not to be supported


m45t3r:
I don't even know if this is a real issue (I don't have a test case to, well, test), however looking at the source code (that is really clean, this is why I noticed it) it seems you can have ``#``, or Unix-style, comments in blank lines, however not as a line comment (in this case, only ``;`` seems to be supported). 

Why? Because of this line: https://github.com/benhoyt/inih/blob/master/ini.c#L50. You only compare the current char with ``;`` and not ``#``, different from this line https://github.com/benhoyt/inih/blob/master/ini.c#L106.

benhoyt:
I know it's a little bit weird, but this is by design. `inih` is compatible with Python's [ConfigParser](https://docs.python.org/2/library/configparser.html) module, which does it this same way -- `#` is only allowed at start of line. This is probably because of compatibility with Windows .INI file parsing, which doesn't recognize `#` at all. So closing this.

m45t3r:
Looking at the last Python (3.5) the behavior changed: https://docs.python.org/3.5/library/configparser.html#customizing-parser-behaviour (``inline_comment_prefixes``). This change seems to be from Python 3.2 onwards.

I actually thing the idea of disabling inline comments a great idea, seems to be more consistent than only allowing ``;`` and not ``#`` for inline comments. It should be simple to add support as a compile time to disable inline comments (just #ifdef the code from ``find_char_or_comment`` and leave only the return).

You can even leave this disabled by default, and I don't think it is worth it to add the option to configure the behavior of inline comments (I think this should be pretty hard to do in C using only preprocessors). However the possibility of disabling inline comments seems great to me.

Need to open another issue?

benhoyt:
Huh, interesting -- good find. I kinda like the idea of adding a preprocessor option to disable inline comments completely.

benhoyt:
This is now done in https://github.com/benhoyt/inih/releases/tag/r34 - you can easily disable inline comments with `-DDINI_ALLOW_INLINE_COMMENTS=0`, or enable handling of both comment characters with `-DINI_INLINE_COMMENT_PREFIXES=";#"`.

benhoyt:
Oops, make that [r35](https://github.com/benhoyt/inih/releases/tag/r35) -- there was a small bug in r34 where it always override your `INI_INLINE_COMMENT_PREFIXES`.

PR #46, opened 2016-01-02
fix parsing '#' comments in the middle of a line

M1cha:


benhoyt:
Per https://github.com/benhoyt/inih/issues/40, I'm actually about to disable inline comments by default (you'll be able to enable them with -DINI_ALLOW_INLINE_COMMENTS, which will enable the behaviour of your fix). This is in line with the current (Python 3.2+) defaults, and the handling of inline comments is a bit weird anyway -- probably why Python changed their defaults, see [here](https://docs.python.org/3/whatsnew/3.2.html#configparser).

Sound okay?

benhoyt:
Actually, on seconds thoughts, I think I'll leave inline comments with only ';' as the default to preserve `inih` backwards compatibility behaviour. Users will still be able to easily disable with `-DDINI_ALLOW_INLINE_COMMENTS=)`, or enable your patch's behaviour with `-DINI_INLINE_COMMENT_PREFIXES=";#"`.

benhoyt:
This is now done in release [r34](https://github.com/benhoyt/inih/releases/tag/r34).

benhoyt:
Oops, make that [r35](https://github.com/benhoyt/inih/releases/tag/r35) -- there was a small bug in r34 where it always override your `INI_INLINE_COMMENT_PREFIXES`.

M1cha:
thx :)
btw shouldn't all parsing related defines be replaced with runtime configuration so u could parse different files with different behavior?

benhoyt:
Well, if this was a full-featured inih library, yes, but inih is "simple, good for embedded systems", so all the settings are global and compile time to keep things simple and the runtime footprint tiny.

PR #47, opened 2016-04-10
Memory usage optimization

artemShelest:
Use references instead of copying strings. INIReader object is actually const afer construction, so I reflected this in the code as well.

### DIFF
--- a/cpp/INIReader.cpp
+++ b/cpp/INIReader.cpp
@@ -18,20 +18,20 @@ INIReader::INIReader(const std::string &filename)
     _error = ini_parse(filename.c_str(), ValueHandler, this);
 }
 
-int INIReader::ParseError()
+int INIReader::ParseError() const
 {
     return _error;
 }
 
-std::string INIReader::Get(const std::string &section, const std::string &name, const std::string &default_value)
+const std::string& INIReader::Get(const std::string &section, const std::string &name, const std::string &default_value) const
 {
     std::string key = MakeKey(section, name);
-    return _values.count(key) ? _values[key] : default_value;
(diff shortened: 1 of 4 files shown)


benhoyt:
Thanks for taking the time to write a PR. Someone actually submitted something very similar to this, and the short version is I'm not sure it's an improvement or that I want to work on the C++ wrapper of inih; I'd rather focus on the C version (I know C well enough to maintain effectively :-). Please read the commentary at https://github.com/benhoyt/inih/pull/42 and let me know if you have further thoughts.

artemShelest:
Hi,
Thank you for your reply.
I believe that modern C++ compilers are capable to make all these
optimizations that have been done in this PR.
I was surprised to see stl strings instead of const char* in this lib. For
my application, using std::string is ok, but for the sake of simplicity I'd
rewrite the whole c++ stuff in this lib using const char*.
The one thing that is beyond #42 is const-correctness - it should be
maintained, regardless of having or not having refs.


2016-04-16 19:15 GMT+03:00 Ben Hoyt <notifications@github.com>:

benhoyt:
Well, `const char*` is not safe, idiomatic C++ -- if you want the `const char*` version, just use the inih C library.

In any case, I agree with your assessment above about const correctness -- I'm going to close this PR, but if you want to submit a new one with just the `const`s after the function definitions for ParseError and the Get* functions, I'd be happy to merge that.

artemShelest:
Agree on the safe code. New PR is out.
Thank you.

2016-04-17 4:06 GMT+03:00 Ben Hoyt <notifications@github.com>:

PR #48, opened 2016-04-17
Make INIReader const correct.

artemShelest:
Made appropriate functions const at INIReader class.


benhoyt:
Looks good to me, thanks! Merged.

PR #50, opened 2016-06-15
Update INIReader.cpp file

jinstrong:
added an validated solution to replace std::map::at method, which is not available until c++11.

### DIFF
--- a/cpp/INIReader.cpp
+++ b/cpp/INIReader.cpp
@@ -26,7 +26,9 @@ int INIReader::ParseError() const
 string INIReader::Get(string section, string name, string default_value) const
 {
     string key = MakeKey(section, name);
-    return _values.count(key) ? _values.at(key) : default_value;
+    return _values.count(key) ? _values.at(key) : default_value; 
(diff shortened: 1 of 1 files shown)


cpp/INIReader.cpp
```diff
@@ -26,7 +26,9 @@ int INIReader::ParseError() const
 string INIReader::Get(string section, string name, string default_value) const
 {
     string key = MakeKey(section, name);
-    return _values.count(key) ? _values.at(key) : default_value;
+    return _values.count(key) ? _values.at(key) : default_value; 
+    //Above statement used std::map::at method, which is not availalbe until c++11. Modify above sentence to below if using an older version of compiler:
+    //return _values.count(key) ? _values.find(key)->second : default_value;
```
benhoyt: Thanks. However, I think we should just clean up the code and use `_values.find(key)->second` always, as it will be compatible with all versions. Can you please change to this:

    // Use _values.find() here instead of _values.at() to support pre C++11 compilers
    return _values.count(key) ? _values.find(key)->second : default_value;

benhoyt:
Looks good to me now, thanks!

PR #53, opened 2016-09-10
[INIReader] class now using constant reference as method arguments.

TheVice:
Hello.

I propose to change argument types from direct type to constant reference so it will be good from performance side as string do not copy at method call.
Get method also can return constant reference so no temporary string will not be provided, it copy already to string that expected to take data from Get method.

Thank you.

P.S.
By the way, if you interesting about other change that can be provide for this class (most expected that compiler support C++ 11) I save my view at [separate branch](https://github.com/TheVice/inih/tree/vicemaster) at repository fork.

### DIFF
--- a/cpp/INIReader.cpp
+++ b/cpp/INIReader.cpp
@@ -23,7 +23,7 @@ int INIReader::ParseError() const
     return _error;
 }
 
-const string& INIReader::Get(const string& section, const string& name, const string& default_value) const
+string INIReader::Get(const string& section, const string& name, const string& default_value) const
 {
     string key = MakeKey(section, name);
     // Use _values.find() here instead of _values.at() to support pre C++11 compilers
--- a/cpp/INIReader.cpp
+++ b/cpp/INIReader.cpp
@@ -13,7 +13,7 @@
 
 using std::string;
 
-INIReader::INIReader(string filename)
+INIReader::INIReader(const string& filename)
 {
     _error = ini_parse(filename.c_str(), ValueHandler, this);
 }
@@ -23,14 +23,14 @@ int INIReader::ParseError() const
     return _error;
 }
 
-string INIReader::Get(string section, string name, string default_value) const
+const string& INIReader::Get(const string& section, const string& name, const string& default_value) const
 {
     string key = MakeKey(section, name);
     // Use _values.find() here instead of _values.at() to support pre C++11 compilers
     return _values.count(key) ? _values.find(key)->second : default_value;
 }
 
-long INIReader::GetInteger(string section, string name, long default_value) const
+long INIReader::GetInteger(const string& section, const string& name, long default_value) const
 {
     string valstr = Get(section, name, "");
     const char* value = valstr.c_str();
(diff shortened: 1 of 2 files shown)


benhoyt:
Hi @TheVice, this has been proposed by several folks (see PR #53 for example), and I'm a bit hesitant for a couple of reasons: 1) There seems to be debate about whether it is good practice and even whether it would help performance with compiler optimizations turned on, and 2) I haven't used C++ in quite a while (particularly not C++11 or 14) and am not in a very good position to review the PR.

However, on #1 if you could do a benchmark comparing speed between the current version and the const ref version (with compiler optimizations turned on), that would be helpful. On #2, will changing the signatures to const ref cause any backwards compatibility issues, or is it fully a drop-in replacement in existing code?

TheVice:
Hi @benhoyt, thank you for your answer.

1) I push [new branch](https://github.com/TheVice/inih/tree/benchmark) that consist primitive benchmark test.
On Linux I got next output after 3 times run of both variants (exist in master and proposer in this pull request):

```
inih/extra$ ./make_benchmark.sh 
inih/extra$ ./INIReaderTest
1.18089 seconds (1180894018 nanoseconds)
inih/extra$ ./INIReaderTest_proposed 
0.964911 seconds (964911319 nanoseconds)
inih/extra$ ./INIReaderTest
1.15753 seconds (1157528693 nanoseconds)
inih/extra$ ./INIReaderTest_proposed 
0.988372 seconds (988371977 nanoseconds)
inih/extra$ ./INIReaderTest
1.2012 seconds (1201201711 nanoseconds)
inih/extra$ ./INIReaderTest_proposed 
0.982808 seconds (982808445 nanoseconds)
```

If change on your machine not impassive you can add zero to [COUNT](https://github.com/TheVice/inih/blob/benchmark/cpp/INIReaderTest.cpp#L14)
```
inih/extra$ ./INIReaderTest_proposed 
9.23079 seconds (9230791354 nanoseconds)
inih/extra$ ./INIReaderTest
11.4242 seconds (11424216960 nanoseconds)
```

By the way, in this pull request I do not propose C++ 11/14 code. Pre C++ 11 compile should still compile this class.

2) Hard to create sample that can be break after proposed change.
Exist using in INIReader do not change after this change.
But in some exotic using it may break, for example if some using specific optimization that expect that there is no reference type at arguments.

just6979:
The const ref on the inputs is a good practice, since inih doesn't need its own copy of the input string and also won't be changing it.

However, the const ref on the returned string is not good, since we probably want the calling app to "own" the returned string, and be able to manipulate it if needed. With a reference, inih "owns" it, and we have to rely on inih to maintain the memory for it, which is not ideal (limits potential internal changes, breaks encapsulation)

TheVice:
Hi @just6979, thank you for join to this conversation.

I agree with you about return and constant reference, new [commit](https://github.com/benhoyt/inih/pull/53/commits/c4c1f31b9de64bea7efc5aa6d9bd2dde8b5811d6) already available that fix that.

About reference parameters @benhoyt notice that similar request present for this project not [once](https://github.com/benhoyt/inih/issues/31), not [twice](https://github.com/benhoyt/inih/pull/42) and may be event not [thrice](https://github.com/benhoyt/inih/pull/47).

Even [benchmark](https://github.com/TheVice/inih/tree/benchmark) show that it must be huge data to see some difference between two approach.

benhoyt:
Thanks -- I'll merge this and create a new release shortly. It does seem best practice, and it doesn't change the calling code, so seems like a win. (Note that in practice the I/O is almost certainly going to be the bottleneck.)

benhoyt:
This is in release https://github.com/benhoyt/inih/releases/tag/r37

PR #54, opened 2016-09-14
Move INIReaderTest.cpp to tests for consistency.

just6979:
Just a tiny change to move the C++ tests to the tests directory with the others


benhoyt:
Hi @just6979, thanks for that. However, that `INIReaderTest.cpp` is somewhat mis-named -- it's really an example, kind of the C++ version of examples/ini_example.c. If we're going to move it, can you please update the pull request to move it to examples/INIReaderExample.cpp?

just6979:
Updated as specified.

benhoyt:
Merged, thanks!

PR #56, opened 2016-12-29
Support for reporting current line

MaG21:
Add new functions that reports the current line in each pass.
without disrupting the nice interface we all love.

newly added functions:
    - ini_parse_stream_report_lines
    - ini_parse_file_report_lines
    - ini_parse_report_lines

This new functionality will allow developers to notify users the exact
place where they provided an unrecognized key and/or value within
the .ini file (which may or may not be a corrupt file)

### DIFF
--- a/examples/ini_dump_report_lines.c
+++ b/examples/ini_dump_report_lines.c
@@ -1,0 +1,40 @@
+/* ini.h example that simply dumps an INI file without comments */
+
+#include <stdio.h>
+#include <string.h>
+#include "../ini.h"
+
+static int dumper(void* user, const char* section, const char* name,
+                  const char* value, int line)
+{
+    static char prev_section[50] = "";
+
+    if (strcmp(section, prev_section)) {
+        printf("%s[%s]\n", (prev_section[0] ? "\n" : ""), section);
+        strncpy(prev_section, section, sizeof(prev_section));
+        prev_section[sizeof(prev_section) - 1] = '\0';
+    }
+    printf("%d) %s = %s\n", line, name, value);
+    return 1;
+}
+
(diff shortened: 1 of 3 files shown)


MaG21:
See:  examples/ini_dump_report_lines.c 

Ei:
```
examples $ ./ini_dump_report_lines test.ini
[protocol]


4) version = 6



[user]
7) name = Bob Smith
8) email = bob@smith.com
9) active = true
10) pi = 3.14159
```

MaG21:
I really like the way you implemented this parser, I did my best not to prostituting your job and yet I wouldn't take it personal if you reject this PR, it feels a bit invasive.

One thing though, IMHO I believe if you keep using the model of using compile time options to add/remove functionality from the parser will eventually bite you, because of the complexity added by those ``#ifdef``.

benhoyt:
Hi, @MaG21 thanks for the pull request. I like the idea -- I'll take a closer look in the next few days.

benhoyt:
As mentioned, I like the feature, however I took a look and your approach -- while clever -- does look a bit invasive for this library. Because inih is designed for (sometimes tiny) embedded systems, I want as little overhead as possible if you don't use a feature, and your solution introduces a code size / runtime penalty.

So I solved this using another `#define INI_HANDLER_LINENO` in https://github.com/benhoyt/inih/commit/30a85920d1260696a307e5c74c660635c00e5992.

I definitely hear you about too many compile-time options making the code get out of hand, but I think how I've done this with the `#define HANDLER` macro in `ini.c` makes the code very little harder to understand.

Thanks again for the feature request! I'll close this PR and create a release tag shortly.

MaG21:
Awesome!

🎉

PR #58, opened 2017-07-09
Add ini_parse_unicode

WilliamFeely:
Adds a version of the ini_parse function that loads a file from a Unicode path.
Since not all embedded platforms might support Unicode, ini.h has added an option INI_ENABLE_UNICODE which can be set to 0 to disable compiling the ini_parse_unicode function.


WilliamFeely:
You may squash this when merging if you desire.

WilliamFeely:
You may squash this when merging if you desire.

benhoyt:
Appreciate the effort. However, I think I'd rather handle this by just letting the user call `ini_parse_file()`, rather than including Unicode / `wchar_t` support in inih, for two reasons:

1) `_wfopen` is a non-portable Windows function
2) If inih starts supporting Unicode / wchar_t for filenames, where do I stop? I don't want the library to become complex by starting to include wchar_t versions of everything.

WilliamFeely:
So the better idea is to call _wfopen from the application then call ini_parse_file() from the application?
Thanks for the feedback.

benhoyt:
Yep, exactly -- thanks!

WilliamFeely:
Just tried it, worked perfect.
This might be something worth noting in the README.md file though, as Windows NT platforms use UTF-16 paths natively.

benhoyt:
Good idea -- added in a "Platform-specific notes" section at the bottom of the readme: https://github.com/benhoyt/inih#platform-specific-notes

PR #59, opened 2017-09-28
Add CMakeLists.txt

yangfl:


benhoyt:
I appreciate the contribution, but I don't really want to be in the business of supporting various build and make tools inside the inih project. I'd rather leave that for the person using inih, as setups and project vary so wildly in C / C++.

yangfl:
This is just a Makefile that install shared library into the system, not (and cant't be) a support for a specific build system. I'm trying to pack this project into Debian, and any project without a Makefile can be very confusing for every end user.

PR #61, opened 2017-12-13
Add support for unlimited lines

nextl00p:


benhoyt:
Thanks for the pull request. As this library is partly intended for small embedded systems where you want to know exactly how much memory you're allocating, I'm not sure about it. What kind of system are you running inih on, and what is the maximum line length you need to support?

nextl00p:
We run it on RTEMS on our [grisp](https://grisp.org) boards.

Erlang boot arguments can get quite long. So we do not want any limit on it.
When a line is to long for the memory, `realloc` should fail. I think also many text editors crash on VERY big files.

There is still the option to have static memory assignment with `INI_USE_STACK 1`.

benhoyt:
Thanks for the details. So does the GRiSP have 384KB RAM (internal), or 64MB (external) available to you? What lengths are we talking about? To me it seems like you could just set it to a reasonable value (4KB, 10KB) and be done with it.

That said, I'd accept this basic approach if it was done slightly differently: an `INI_ALLOW_REALLOC` option (defaults to off). That way you can have heap allocation but get control over whether it reallocs are not.

nextl00p:
We parse the .ini file in the bootloader, in this case it's internal RAM and also after startup, then we have 64MB of RAM.

I added now some more options. What do you think? ;-)

benhoyt:
Thanks. I'll take a closer look and then merge. However, note that it may not be till January when I have a block of time for this.

benhoyt:
Thanks for your work on this. I implemented and merged your suggestion with a couple of changes:

* It no longer supports unlimited size. You have to set `INI_MAX_LINE` to an upper bound (but it can be huge). I found this just added complication to the (already-complicated) options, and also slightly complicated the code.
* Fixed a bug where if the size doubled to just over the maximum, it wouldn't even allocate the rest of `INI_MAX_LINE` bytes. For example, if the initial size was 600 and INI_MAX_LINE was 1000, it wouldn't realloc. Now it reallocs to exactly the max (1000 bytes).
* Don't do a `strlen()` on the line from the start every loop. This is O(N^2) behaviour. Only count the characters from the current offset onwards.
* Slight code and doc cleanup to fit existing style.
* Made two sections for describing options in README.

Please review the change in https://github.com/benhoyt/inih/commit/d709bdac5b6bfa27c913d3a3e0e982d2c12217f2 and let me know if you have any feedback / concerns.

PR #63, opened 2018-04-02
Add support for allowing name without value

Contextualist:
Per Python configparser's [init option](https://docs.python.org/3/library/configparser.html#customizing-parser-behaviour) `allow_no_value`, which makes the following a legitimate config:
```ini
a name without value
```
with name `"a name without value"` and value `NULL`.

### DIFF
--- a/ini.c
+++ b/ini.c
@@ -192,9 +192,23 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
                 if (!HANDLER(user, section, name, value) && !error)
                     error = lineno;
             }
-            else if (!error) {
+            else {
                 /* No '=' or ':' found on name[=:]value line */
-                error = lineno;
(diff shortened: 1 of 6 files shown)


benhoyt:
Thanks for the PR. However, this doesn't seem particularly useful to me, as many INI parsers don't support it, and you can just add an `= 1` or `= true` to your config file. Is your use case to support existing "allow_no_value" INI files, and are you able to edit the config files instead? I'm just wary of adding more options and `#ifdefs`.

Contextualist:
Fine, then let's keep the implementation minimal. After all, this option belongs to the extended standard. I use your lib for my personal project (so I actually dictate the format of the config file :) mostly because I appreciate its simplicity 👍 . I favor this option just to avoid redundancy of adding a value for the set-like part of a config.

benhoyt:
Sounds good to me, thanks :-)

PR #65, opened 2018-07-03
Fixed cppcheck warnings

FredrikBlix:
I ran cppcheck on the code and found two warnings, fixed them.

### DIFF
--- a/cpp/INIReader.cpp
+++ b/cpp/INIReader.cpp
@@ -73,7 +73,7 @@ string INIReader::MakeKey(const string& section, const string& name)
 int INIReader::ValueHandler(void* user, const char* section, const char* name,
                             const char* value)
 {
-    INIReader* reader = (INIReader*)user;
+    INIReader *reader = static_cast<INIReader*>(user);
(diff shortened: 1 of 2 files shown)


cpp/INIReader.cpp
```diff
@@ -73,7 +73,7 @@ string INIReader::MakeKey(const string& section, const string& name)
 int INIReader::ValueHandler(void* user, const char* section, const char* name,
                             const char* value)
 {
-    INIReader* reader = (INIReader*)user;
+    INIReader *reader = static_cast<INIReader*>(user);
```
benhoyt: Thanks -- happy to merge this in. Just one minor nit-pick before I do: can you please change `INIReader *reader ...` back to `INIReader* reader ...` with the space after the `*`? That would be consistent with the rest of the file and it's also more typical C++ style.

FredrikBlix: Fixed! Sorry, my bad...

benhoyt:
Thanks!

PR #67, opened 2018-07-22
Avoid cppcheck free() error

odgalvin:
This avoids this error and warning in cppcheck:
```
[ini.c:92]: (style) The scope of the variable 'offset' can be reduced.
[ini.c:127]: (error) Deallocation of an auto-variable results in undefined behaviour.
```

We don't need to free 'line' if we're using the stack. This also avoids problems if the user misconfigures their inih options by setting INI_ALLOW_REALLOC while using the stack, which as ini.h says "Only applies if INI_USE_STACK is zero."


benhoyt:
Thanks!

PR #69, opened 2018-09-04
Add HasValue method to INIReader

Jvanrhijn:
This just adds a method to check whether a given field exists, which I've found to be useful in a project of mine. 

Thanks for your great work!

### DIFF
--- a/cpp/INIReader.h
+++ b/cpp/INIReader.h
@@ -42,6 +42,9 @@ public:
     // and valid false values are "false", "no", "off", "0" (not case sensitive).
     bool GetBoolean(const std::string& section, const std::string& name, bool default_value) const;
 
+    // Checks whether the key at the given section exists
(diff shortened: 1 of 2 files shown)


cpp/INIReader.h
```diff
@@ -42,6 +42,9 @@ class INIReader
     // and valid false values are "false", "no", "off", "0" (not case sensitive).
     bool GetBoolean(const std::string& section, const std::string& name, bool default_value) const;
 
+    // Checks whether the key at the given section exists
```
benhoyt: Let's make this comment a bit more accurate:

    // Return true if a value exists with the given section and field names.

benhoyt:
In general I'm not adding to / maintaining the C++ API, but this is a trivial addition and I'm in favor. Just one comment about the comment. :-)

PR #72, opened 2018-10-10
Add INIReader::GetString to get nicer default_value behavior

NeatNit:
Same as `INIReader::Get`, but returns `default_value` if the value is an empty string, which provides functionality similar to `GetInteger` and the others.

I'm not sure about the function name (implies that `Get` is something else?), but I definitely don't want to change the behavior of `Get` and I also think this functionality is really needed. `GetString` is the best name I could come up with that isn't overly convoluted (e.g. I ruled out `GetButIfValueIsEmptyStringThenReturnDefaultValueInstead`) and I don't think it'll cause anyone trouble.

### DIFF
--- a/cpp/INIReader.cpp
+++ b/cpp/INIReader.cpp
@@ -30,6 +30,12 @@ string INIReader::Get(const string& section, const string& name, const string& d
     return _values.count(key) ? _values.find(key)->second : default_value;
 }
 
+string INIReader::GetString(const string& section, const string& name, const string& default_value) const
+{
+    const string str = Get(section, name, "");
+    return str.empty() ? default_value : str;
+}
+
 long INIReader::GetInteger(const string& section, const string& name, long default_value) const
 {
     string valstr = Get(section, name, "");
--- a/cpp/INIReader.h
+++ b/cpp/INIReader.h
@@ -27,6 +27,10 @@ public:
     // Get a string value from INI file, returning default_value if not found.
     std::string Get(const std::string& section, const std::string& name,
                     const std::string& default_value) const;
+    
+    // Get a string value from INI file, returning default_value if not found or empty.
+    std::string GetString(const std::string& section, const std::string& name,
+                    const std::string& default_value) const;
 
     // Get an integer (long) value from INI file, returning default_value if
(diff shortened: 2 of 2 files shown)


benhoyt:
Hi @NeatNit, thanks for the PR. I'm okay with this, but I don't think the name `GetString` fits 100%, as the other `GetInteger` etc have slightly different semantics. They return the default if the field is not present, but this will return it if it's not found or empty. So how about `GetNoEmpty`?

Also, I think it'd be good to clarify that (because `ini.c` trims whitespace at the start and end of the value), that it will return the default if the value is empty or contains only whitespace characters.

NeatNit:
In practice, `GetInteger`, `GetReal` and `GetBoolean` all return `default_value` for empty settings, as a side effect of the way they're written. Here's the code with comments added to clarify:

```CPP
long INIReader::GetInteger(const string& section, const string& name, long default_value) const
{
    string valstr = Get(section, name, "");
    const char* value = valstr.c_str();
    char* end;
    // This parses "1234" (decimal) and also "0x4D2" (hex)
    long n = strtol(value, &end, 0);
    // if valstr was empty, end = value, and so default_value is returned:
}

double INIReader::GetReal(const string& section, const string& name, double default_value) const
{
    string valstr = Get(section, name, "");
    const char* value = valstr.c_str();
    char* end;
    double n = strtod(value, &end);
    // if valstr was empty, end = value (I'm pretty sure), and so default_value is returned:
}

bool INIReader::GetBoolean(const string& section, const string& name, bool default_value) const
{
    string valstr = Get(section, name, "");
    // Convert to lower case to make string comparisons case-insensitive
    std::transform(valstr.begin(), valstr.end(), valstr.begin(), ::tolower);
    // if valstr was empty, none of these match and else is reached, returning default_value:
    if (valstr == "true" || valstr == "yes" || valstr == "on" || valstr == "1")
        return true;
    else if (valstr == "false" || valstr == "no" || valstr == "off" || valstr == "0")
        return false;
    else
        return default_value;
}
```

So `GetString` would match this behavior.

If you still want me to change the name, I'll be happy to do that, I just wanted to make sure you were aware of this :)

NeatNit:
Is this what you meant?

benhoyt:
Looks good to me, thanks!

PR #76, opened 2019-02-28
Allow handler to receive section number

fmor:
Hi, 
This change allow to distingish multiples sections with same name and so build an array.

# conf.ini
[sectionA]
param=value1

[sectionA]
param=value2

# handler.c
typdef struct
{
    int a;
} Element;

typedef struct 
{
    Element* elements;    
    int count;
} Array;

int  _inihandler(void* user, const char* section, const char* name, const char* value,int sectionno)
{

    Array* array;
    Element* element;

    array = (Array*) user;

    if( sectiono != array->count )
    {
        array->elements = realloc( array->elements, sizeof(Element) * sectionno );
        element = array->elements[array->count];
        array->count++;
    }
}

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ You can control various aspects of inih using preprocessor defines:
 
   * **Stop on first error:** By default, inih keeps parsing the rest of the file after an error. To stop parsing on the first error, add `-DINI_STOP_ON_FIRST_ERROR=1`.
   * **Report line numbers:** By default, the `ini_handler` callback doesn't receive the line number as a parameter. If you need that, add `-DINI_HANDLER_LINENO=1`.
+  * **Report section number:** By default, the `ini_handler` callback doesn' receive the section number as a parameter. If you need that add `-DINI_HANDLER_SECTIONNO=1`, INI_HANDLER_LINENO and INI_HANDLER_SECIONNO are mutualy exclusives and INI_HANDLER_LINENO is checked first.
 
 ### Memory options ###
 
--- a/ini.c
+++ b/ini.c
@@ -100,6 +100,7 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
     char* value;
     int lineno = 0;
     int error = 0;
+    int sectionno = 0;
 
 #if !INI_USE_STACK
     line = (char*)malloc(INI_INITIAL_ALLOC);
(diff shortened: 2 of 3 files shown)


benhoyt:
Hi @fmor, thanks for the suggestion. However, I don't understand why a new feature is needed for this? Can't you just append to the array if the section+name is repeated? You don't even need (and it's probably not typical) to use a new section, just a repeated key name. I've seen a couple of ways of doing this, one is just repeating the key:

```
[settings]
email = one@example.com
email = two@example.com
```

And the section is using a `[]` name suffix to specifically indicate it's an array:

```
[settings]
emails[] = one@example.com
emails[] = two@example.com
```

I kind of like the second one, as it's explicit and you could support it in dynamically typed languages easily as well, where you may not know the types of the settings beforehand.

To parse that you could do something like this (not tested!):

```c
typedef struct {
    char** elements;
    int count;
} Array;

int handler(void* user, const char* section, const char* name, const char* value) {
    Array* array;
    Element* element;
    array = (Array*)user;

    if (strcmp(section, "settings")==0 && strcmp(name, "emails[]")==0) {
        array->elements = realloc(array->elements, sizeof(char*) * (array->count + 1));
        array->elements[array->count] = strdup(value);
        array->count++;
    }
}
```

fmor:
Thanks for your quick reply, 

If a param is absent and then should take a default value, the array version force end user to write the all serie with default values, it could be indexed but as long as elements and params  count increase, ini file become complex and then could induce errors. 

With array :

[users]
mails[]=johndoe@
mails[]=fpignon@
names[]=john doe
names[]=francois pignon
send_mail[]=0
send_mail[]=1
send_candy[]=1
send_candy[]=0
arrays[]=1,2,3,4,5
arrays[]=0

With section :

[user]
name=john doe
mail=jdoe@
send_candy=1
array=1,2,3,4,5

[user]
name=francois pignon
mail=fpignon@
send_mail=1

benhoyt:
Unfortunately I'd rather not merge this feature. Partly it feels like feature creep, but also because INI many INI parsers don't recognize duplicate section names as valid. Specifically, Python's [configparser](https://docs.python.org/3/library/configparser.html), whose behavior `inih` is modelled on, doesn't support it (and even raises an error in the default `strict` mode).

INI format is really for very flat, two-level structures. So you're probably better off using something like TOML, or YAML, or JSON for arrays of objects like this. Alternatively you could use different section names like `[user1]`, `[user2]`. Or increment the section number whenever you see the `name` field (though that would make it order-dependent).

fmor:
[userX] 
Well.... I think i was a little bit too focused to think of that.
Thank you

PR #78, opened 2019-03-18
C++17, VS2017 project

DBJDBJ:
Complete reshuffle 
Complete rewrite of the C++ part.


benhoyt:
Hi @DBJDBJ, but unfortunately I can't accept this contribution. It's a significant and backwards-incompatible change to the C++ API. Additionally, it looks like you've removed all the automated test files. And the MSVC makefiles is something I don't have a Windows machine available to support. My suggestion is that you fork this project, or perhaps even better, simply use inih in your own project and create a better C++ wrapper.

DBJDBJ:
@benhoyt  If you think you would like to use any of it please do, with the help of "copy-paste".  I think it should provide significant performance improvements. In the next release, I will remove the use of `std::string` completely. I shall proceed with my own project. Kind regards...

PR #79, opened 2019-03-22
Add option to call handler when a new section is encountered

ksdhans:
Some programs have ini files with multiple sections that have the same name, which would make it very useful to know when a new section is encountered.

This patch adds this ability as an option that's enabled by setting INI_CALL_HANDLER_ON_NEW_SECTION to 1.

When enabled, the handler callback can check the name parameter. If it's NULL, then it was called because a new section was entered.

### DIFF
--- a/tests/unittest.c
+++ b/tests/unittest.c
@@ -27,19 +27,26 @@ int dumper(void* user, const char* section, const char* name,
 #endif
 {
     User = *((int*)user);
-    if (strcmp(section, Prev_section)) {
+    if (!name || strcmp(section, Prev_section)) {
         printf("... [%s]\n", section);
         strncpy(Prev_section, section, sizeof(Prev_section));
         Prev_section[sizeof(Prev_section) - 1] = '\0';
     }
 
 #if INI_HANDLER_LINENO
-    printf("... %s=%s;  line %d\n", name, value, lineno);
+	if(name) {
+		printf("... %s=%s;  line %d\n", name, value, lineno);
+	} else {
+		/* No name means that this was called for a section header */
+		printf("... ;  line %d\n", lineno);
+	}
 #else
-    printf("... %s=%s;\n", name, value);
+	if(name) {
+		printf("... %s=%s;\n", name, value);
+	}
 #endif
 
-    return strcmp(name, "user")==0 && strcmp(value, "parse_error")==0 ? 0 : 1;
+    return name && strcmp(name, "user")==0 && strcmp(value, "parse_error")==0 ? 0 : 1;
 }
 
 void parse(const char* fname) {
@@ -62,5 +69,6 @@ int main(void)
     parse("multi_line.ini");
     parse("bad_multi.ini");
     parse("bom.ini");
+	parse("duplicate_sections.ini");
     return 0;
 }
--- a/tests/unittest.c
+++ b/tests/unittest.c
@@ -66,6 +66,6 @@ int main(void)
     parse("multi_line.ini");
     parse("bad_multi.ini");
     parse("bom.ini");
-	parse("duplicate_sections.ini");
+    parse("duplicate_sections.ini");
     return 0;
 }
--- a/tests/unittest.c
+++ b/tests/unittest.c
@@ -32,21 +32,18 @@ int dumper(void* user, const char* section, const char* name,
         strncpy(Prev_section, section, sizeof(Prev_section));
         Prev_section[sizeof(Prev_section) - 1] = '\0';
     }
+    if (!name) {
+        return 1;
+    }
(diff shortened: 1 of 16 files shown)


benhoyt:
Similar to the rationale for #76: unfortunately I'd rather not merge this feature. Partly it feels like feature creep, but also because many INI parsers don't recognize duplicate section names as valid. Specifically, Python's [configparser](https://docs.python.org/3/library/configparser.html), whose behavior inih is modelled on, doesn't support it (and even raises an error in the default strict mode).

ksdhans:
I hope you'll reconsider. Existing users (and you) who want configparser like behaviour won't be affected because you have to explicitly enable the feature. Meanwhile, some of us need to be able to process INI files with the same name. And no, numbering those sections is NOT a good solution with user-editable INI files; it increases the chance of mistakes being made, and can give confusing behaviour since the parser doesn't give a warning when it happens and there's no reliable way to detect consecutive sections with the same name (i.e., I'd get bug reports from frustrated users claiming that my config file loading is broken).

Anyway, it's your call.

benhoyt:
@ksdhans Yep, fair points -- reconsidering!

I'm happy including this feature, and I think it's a better approach than the numbering approach. However, it needs tests and polish. Specifically you'll need to add:

* New test command lines in `tests/unittest.bat` and `tests/unittest.sh`
* New test ini files if applicable (probably can just use the existing ones and turn on the option)
* Baseline test file(s)
* Documentation in `README.md` under "Parsing options"
* Documentation of the flag in `ini.h`

ksdhans:
Okay, I'll have a look at adding the tests and docs either this weekend or next week.

ksdhans:
I've added the documentation and tests. You may wish to double-check that tests/unittest.bat works as it should, because I don't have tcc. The tests worked fine via tests/unittest.sh in cygwin, though.

tests/unittest.c
```diff
@@ -27,19 +27,26 @@ int dumper(void* user, const char* section, const char* name,
 #endif
 {
     User = *((int*)user);
-    if (strcmp(section, Prev_section)) {
+    if (!name || strcmp(section, Prev_section)) {
```
benhoyt: I'd prefer if we got rid of the changes below, and short-circuited early after this section. For example:

```c
    if (!name || strcmp(section, Prev_section)) {
        printf("... [%s]\n", section);
        strncpy(Prev_section, section, sizeof(Prev_section));
        Prev_section[sizeof(Prev_section) - 1] = '\0';
    }
    if (!name) {
        return 1;
    }
```

That will keep the code significantly simpler.

tests/unittest.c
```diff
@@ -27,19 +27,26 @@ int dumper(void* user, const char* section, const char* name,
...
         strncpy(Prev_section, section, sizeof(Prev_section));
         Prev_section[sizeof(Prev_section) - 1] = '\0';
     }
 
 #if INI_HANDLER_LINENO
-    printf("... %s=%s;  line %d\n", name, value, lineno);
+	if(name) {
+		printf("... %s=%s;  line %d\n", name, value, lineno);
+	} else {
+		/* No name means that this was called for a section header */
```
benhoyt: I don't think this section is used (and will go away if we do the above anyway).

tests/unittest.c
```diff
@@ -27,19 +27,26 @@ int dumper(void* user, const char* section, const char* name,
...
+	}
 #else
-    printf("... %s=%s;\n", name, value);
+	if(name) {
+		printf("... %s=%s;\n", name, value);
+	}
 #endif
 
-    return strcmp(name, "user")==0 && strcmp(value, "parse_error")==0 ? 0 : 1;
+    return name && strcmp(name, "user")==0 && strcmp(value, "parse_error")==0 ? 0 : 1;
```
benhoyt: `name &&` will no longer be needed if the above is done.

tests/unittest.c
```diff
@@ -62,5 +69,6 @@ int main(void)
     parse("multi_line.ini");
     parse("bad_multi.ini");
     parse("bom.ini");
+	parse("duplicate_sections.ini");
```
benhoyt: Should be 4 spaces instead of a tab for consistency.

ksdhans:
As requested, I've simplified the code and fixed the indentation.

benhoyt:
Thanks @ksdhans -- I've merged this now. I'll prepare a versioned release in the next few days.

benhoyt:
Published in release r43. Also added TravisCI testing.

ksdhans:
Thanks. I've deleted my fork, so people don't accidentally use that one instead of yours.

PR #81, opened 2019-05-19
Add HasSection to C++ API

evorw:
This function lets you check for the existence of a section, similar to HasValue. We found this useful, hopefully others do too.

### DIFF
--- a/cpp/INIReader.cpp
+++ b/cpp/INIReader.cpp
@@ -69,7 +69,7 @@ bool INIReader::GetBoolean(const string& section, const string& name, bool defau
 }
 
 // Return true if the given section exists
-bool INIReader::HasSection(const std::string& section) const
+bool INIReader::HasSection(const string& section) const
 {
     const string test = MakeKey(section, "");
     return std::any_of(_values.begin(), _values.end(),
--- a/cpp/INIReader.cpp
+++ b/cpp/INIReader.cpp
@@ -71,12 +71,12 @@ bool INIReader::GetBoolean(const string& section, const string& name, bool defau
 // Return true if the given section exists
 bool INIReader::HasSection(const string& section) const
 {
-  const string key = MakeKey(section, "");
-  std::map<string, string>::const_iterator pos = _values.lower_bound(key);
-  if (pos == _values.end())
-      return false;
-  // Does the key at the lower_bound pos start with "section"?
-  return pos->first.compare(0, key.length(), key) == 0;
+    const string key = MakeKey(section, "");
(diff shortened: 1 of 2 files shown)


cpp/INIReader.cpp
```diff
@@ -68,6 +68,14 @@ bool INIReader::GetBoolean(const string& section, const string& name, bool defau
         return default_value;
 }
 
+// Return true if the given section exists
```
benhoyt: Let's remove this comment so the doc is just in one place (the header file).

cpp/INIReader.cpp
```diff
@@ -68,6 +68,14 @@ bool INIReader::GetBoolean(const string& section, const string& name, bool defau
...
+bool INIReader::HasSection(const std::string& section) const
```
benhoyt: Let's remove the `std::` prefix here as it's not needed with the `using` at the top (I realize `HasValue` has it, but I should remove that).

```suggestion
bool INIReader::HasSection(const string& section) const
```

cpp/INIReader.cpp
```diff
@@ -68,6 +68,14 @@ bool INIReader::GetBoolean(const string& section, const string& name, bool defau
...
+{
+    const string test = MakeKey(section, "");
+    return std::any_of(_values.begin(), _values.end(),
+        [&](const std::pair<string, string>& k){return k.first.find(test) == 0;});
```
benhoyt: I believe `any_of` loops through all values in the map. This is an ordered map, so we should be able to to a binary search for the position and see if a value with that section exists. How about something like this (not extensively tested):

```c++
bool INIReader::HasSection(const string& section) const
{
    string key = MakeKey(section, "");
    std::map<string, string>::const_iterator pos = _values.lower_bound(key);
    if (pos == _values.end())
        return false;
    // Does the key at the lower_bound pos start with "section"?
    return pos->first.compare(0, key.length(), key) == 0;
}
```

cpp/INIReader.h
```diff
@@ -47,6 +47,9 @@ class INIReader
     // and valid false values are "false", "no", "off", "0" (not case sensitive).
     bool GetBoolean(const std::string& section, const std::string& name, bool default_value) const;
 
+    // Return true if the given section exists
```
benhoyt:
```suggestion
    // Return true if the given section exists (section must contain at least
    // one name=value pair).
```

cpp/INIReader.cpp
evorw: Yep, that's better. I updated the PR.

PR #84, opened 2019-07-11
Free should not be used on stack variable

klama-f:
Using free on a stack variable is undefined behavior that is best avoided


klama-f:
Will submit more complete pull request soon

benhoyt:
Hmm, I don't understand why this is needed. That whole block of code is already inside a `!INI_USE_STACK` block, see:

https://github.com/benhoyt/inih/blob/master/ini.c#L119-L137

klama-f:
Sorry, I didn't notice until soon after closing the pull request again.

PR #85, opened 2019-07-27
2019 07 add copyright and spdx

ametzler:
Afaiui inih is not shipped as a shared library but is designed for copying the relevant files to inih-using project's sources.  Especially for this use-case it is very helpful to have proper attribution and clear, accessible information about the applicable license terms _in each file_.

This merge request adds
a) a copyright statement and
b) an SPDX id  https://spdx.org/ids to have in-file license information.

Thanks for considering!


ametzler:
Hello,

I moved the identifier down (amended and force-pushed). - Is it okay this way?

cu Andreas

PR #87, opened 2019-10-03
Allow handler to be called on a key without '=' or ':'

weltling:
The actual need on this is as follows. The INI file contains a special section containing all the sections in the file. That is used to validate the INI file itself. Eg

```
[list all]
section0
section1

[section0]
key=val

[section1]
key=val
```

For this to work, the keys in the `list all` section need to be passed to the handler, so they can be recorded and used later for the validation. This has been made configurable, so won't affect the default library behavior.

Thanks.


benhoyt:
Hi @weltling, this is reasonable request - however, inih already has a ton of `#if`'s and I'm a bit hesitant to add yet another one. Do you have control over your INI file? If so, could you simply add `=` signs with a blank value so it parses? For example:

```
[list all]
section0=
section1=
```

weltling:
Hi @benhoyt, thanks for checking the PR.

Leaving aside that changing the INI spec for the particular project would likely have to go through some cycles, another part of the issue is that the file is used by different parties and at least one already uses a parser that supports this kind of behavior. I fully understand your point to not to pollute `inih` with `#if`'s, so ofc it's up to you. I'll have to check, whether carrying teh patch would be a solution, or come up with something else. In any case, thanks for your work and a useful piece of software :)

Thanks.

benhoyt:
Cool, let me know what you find and comment back here if you can't work around it.

bluca:
@benhoyt would unconditionally allow value-less keys be an option? That way there's no extra ifdefs. There are precedents in other ini parsers support this.
Thanks!

benhoyt:
Okay - I've looked at this a bit further and I'm reopening. I don't want to enable this by default, but I'm willing to add the `#if` version. Python's configparser (which inih is somewhat modeled after) flags this as an error by default too, but has an `allow_no_value` option to enable this mode. I'll request a couple of tweaks to the PR and the addition of tests -- reviewing now.

ini.c
```diff
@@ -202,7 +202,12 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
             }
             else if (!error) {
                 /* No '=' or ':' found on name[=:]value line */
-                error = lineno;
+#if INI_ALLOW_KEY_WITHOUT_EQUAL
```
benhoyt: Let's call this `INI_ALLOW_NO_VALUE` to match Python's "allow_no_value" option.

ini.c
```diff
@@ -202,7 +202,12 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
...
+                *end = '\0';
+                name = rstrip(start);
+                if (!HANDLER(user, section, name, NULL))
```
benhoyt: You need to add a `&& !error` here like the other `HANDLER` calls so that it returns the first error.

ini.c
```diff
@@ -202,7 +202,12 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
...
+#endif
+			error = lineno;
```
benhoyt: This is clever :-) but the indentation not matching up is a bit weird. Can we do this instead?

```c
#if INI_ALLOW_NO_VALUE
        *end = '\0';
        name = rstrip(start);
        if (!HANDLER(user, section, name, NULL))
            error = lineno;
#else
        error = lineno;
#endif
```

ini.h
```diff
@@ -134,6 +134,13 @@ int ini_parse_string(const char* string, ini_handler handler, void* user);
 #define INI_CALL_HANDLER_ON_NEW_SECTION 0
 #endif
 
+/* If a section contains a key without '=' or ':' signs, it would be not
+   passed to the handler. Setting the below to 1 will still call the handler
+   and explicitly set value to NULL. */
```
benhoyt: Let's change the wording here to match the style of the other ones:

```c
/* Nonzero to allow a name without a value (no '=' or ':' on the line) and
   call the handler with value NULL in this case. Default is to treat
   no-value lines as an error. */
```

weltling:
@benhoyt glad you change your mind :) I'm working on some tests, will update next days.

Thanks.

weltling:
@benhoyt ready for the next round, I think. Please let me know, if something is missing.

Thanks.

tests/unittest.sh
```diff
@@ -51,3 +51,7 @@ rm -f unittest_heap_string
 gcc ../ini.c -DINI_CALL_HANDLER_ON_NEW_SECTION=1 unittest.c -o unittest_call_handler_on_new_section
 ./unittest_call_handler_on_new_section > baseline_call_handler_on_new_section.txt
 rm -f unittest_call_handler_on_new_section
+
+gcc ../ini.c -DINI_ALLOW_NO_VALUE=1 unittest.c -o unittest_call_allow_no_value
+./unittest_call_allow_no_value > baseline_call_allow_no_value.txt
+rm -f unittest_call_allow_no_value
```
benhoyt: This test and the committed `.txt` file shouldn't have the word "call" in it. Should be called just `unittest_allow_no_value`.

tests/unittest.c
```diff
@@ -37,9 +37,17 @@ int dumper(void* user, const char* section, const char* name,
     }
 
 #if INI_HANDLER_LINENO
+# if INI_ALLOW_NO_VALUE
+    printf("... %s%s%s;  line %d\n", name, value ? "=" : "", value ? value : "", lineno);
```
benhoyt: Why don't we simplify this by avoiding the `#if INI_ALLOW_NO_VALUE` and just use the line above in all cases? (Value will never be NULL if INI_ALLOW_NO_VALUE is off anyway.) Same below.

tests/baseline_stop_on_first_error.txt
```diff
@@ -55,3 +55,4 @@ bom.ini: e=0 user=107
 ... single1=def;
 ... single2=qrs;
 duplicate_sections.ini: e=0 user=108
+no_value.ini: e=2 user=108
```
benhoyt: Weird -- I couldn't see immediately why user was not incremented here. Is this a regression?

weltling: Debugged this one. The flow is as follows - it reads the '[section_list]', then it jumps out of `ini_parse` because because the second line doesn't contain a value. The handler is never called. It was in general the behavior as before - the whole section would actually be ignored, if it contained just keys and no values.

In other cases `INI_STOP_ON_FIRST_ERROR` is zero by default, so `ini_parse` goes ahead and finds other sections, so the handler is called. This doesn't seem a regression.

Thanks.

tests/unittest.c
weltling: Changed this one, but GitHub doesn't seem to catch up. Thanks.

weltling:
Everything should be resolved now. Thanks for the quick check.

benhoyt:
Thanks for this -- I'll merge and release a new version shortly.

weltling:
Thanks a lot!

benhoyt:
Released r46 with this feature: https://github.com/benhoyt/inih/releases/tag/r46 -- thanks!

PR #88, opened 2019-10-07
Fix possible int overflow in reader cb

weltling:
This fixes compile failures with the strict compiler flags, please see below.

It's to see, that `(ini_reader)fgets` is actually used as reader, which actually uses int. Still, it is done as an explicit cast. The surrounding errors reported by the compiler are about `strlen` result assigned to an int var, and int var passed to realloc. An enhancement to this might be to additionally write a wrapper around fgets, which would use `size_t` in the signature. 

Another way to solve this would be to keep ints, and do explicit casts. IMO size_t is a better solution, for both type safety and security.

Finally, i'm not sure how bad it is to break the existing API, but from the usability perspective this should be fixed one or another way.

Thanks.

```
inih/ini.c: In function ‘ini_parse_stream’:
inih/ini.c:124:18: error: conversion from ‘size_t’ {aka ‘long unsigned int’} to ‘int’ may change value [-Werror=conversion]
         offset = strlen(line);
                  ^~~~~~
inih/ini.c:129:38: error: conversion to ‘size_t’ {aka ‘long unsigned int’} from ‘int’ may change the sign of the result [-Werror=sign-conversion]
             new_line = realloc(line, max_line);
                                      ^~~~~~~~
inih/ini.c:139:20: error: conversion to ‘size_t’ {aka ‘long unsigned int’} from ‘int’ may change the sign of the result [-Werror=sign-conversion]
             offset += strlen(line + offset);
                    ^~
inih/ini.c:139:23: error: conversion from ‘size_t’ {aka ‘long unsigned int’} to ‘int’ may change value [-Werror=conversion]
             offset += strlen(line + offset);
                       ^~~~~~
```

### DIFF
--- a/ini.c
+++ b/ini.c
@@ -86,14 +86,14 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
     /* Uses a fair bit of stack (use heap instead if you need to) */
 #if INI_USE_STACK
     char line[INI_MAX_LINE];
-    int max_line = INI_MAX_LINE;
+    size_t max_line = INI_MAX_LINE;
 #else
     char* line;
-    int max_line = INI_INITIAL_ALLOC;
+    size_t max_line = INI_INITIAL_ALLOC;
 #endif
 #if INI_ALLOW_REALLOC && !INI_USE_STACK
     char* new_line;
-    int offset;
+    size_t offset;
 #endif
     char section[MAX_SECTION] = "";
     char prev_name[MAX_NAME] = "";
--- a/ini.c
+++ b/ini.c
@@ -86,7 +86,7 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
     /* Uses a fair bit of stack (use heap instead if you need to) */
 #if INI_USE_STACK
     char line[INI_MAX_LINE];
-    size_t max_line = INI_MAX_LINE;
+    int max_line = INI_MAX_LINE;
 #else
     char* line;
     size_t max_line = INI_INITIAL_ALLOC;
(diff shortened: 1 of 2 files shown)


benhoyt:
I'm unsure what to do here -- `fgets` has the len as an int rather than size_t (I guess because it's so old). And I kind of want it to be compatible with `fgets`.

However, I can't reproduce the warnings you're seeing above. For example:

```
$ gcc -Wall examples/ini_dump.c ini.c
~/Home/inih$
```

I'm on a Mac so I think the above "gcc" is actually LLVM. Maybe that's the issue. What command line flags are you using to get these warnings?

weltling:
Thanks for checking. The minimal subset of the flags is 

`gcc -I. -o ini.o -DINI_USE_STACK=0 -DINI_ALLOW_REALLOC=1 -Wconversion -Wall -Werror ini.c`

Reproduced with GCC version 8.2.0 or 8.3.0.

Yeah, `fgets` is quite old, that`s where I've expressed the concern about the API breach. On could go another route to say `offset = (int)strlen(line)` and `(size_t)max_line`. That is ugly, but would allow the signature `fgets` compatibility. 

In the real world, it would be of course difficult to create a malicious ini, especially as `max_line` is predefined. So it's more a compilation issue with certain compiler flags. I'll try to work out another solution for this, to keep the reader signature unchanged, please check if there are other ideas.

Thanks.

benhoyt:
Yeah, I think I'd be hesitant to change the `reader` signature at this point, so it seems like the `(int)strlen()` and `(size_t)max_line` approach is the only real option.

weltling:
Thanks!

PR #92, opened 2019-12-11
INIReader: add constructor accepting a buffer

vrresto:


benhoyt:
Works for me, thanks!

PR #93, opened 2019-12-14
fixed: INIReader::ValueHandler() crashes when value is NULL

vrresto:


benhoyt:
Hi @vrresto, thanks for this! I fixed this this a slightly different way in https://github.com/benhoyt/inih/commit/1e7a61d696d975051c9bf46213b4a8384e0936b6 -- this happens when INI_ALLOW_NO_VALUE is enabled, and it's useful to have the name there with an empty value so you know it's set.

I also fixed a similar issue when `name` is NULL when INI_CALL_HANDLER_ON_NEW_SECTION is enabled. Commit https://github.com/benhoyt/inih/commit/75fe6b1a03d99a9728b9924f9af30729e51357c2

These are in release r47: https://github.com/benhoyt/inih/releases/tag/r47

PR #95, opened 2020-02-01
Freed memory before exiting in examples/ini_example.c

girish946:
Signed-off-by: Girish Joshi <girish946@gmail.com>

### DIFF
--- a/examples/ini_example.c
+++ b/examples/ini_example.c
@@ -40,7 +40,9 @@ int main(int argc, char* argv[])
     }
     printf("Config loaded from 'test.ini': version=%d, name=%s, email=%s\n",
         config.version, config.name, config.email);
-    free((void *) config.name);
-    free((void *) config.email);
+
+    free((void*)config.name);
+    free((void*)config.email);
+
     return 0;
 }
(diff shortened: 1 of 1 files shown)


examples/ini_example.c
```diff
@@ -40,5 +40,7 @@ int main(int argc, char* argv[])
     }
     printf("Config loaded from 'test.ini': version=%d, name=%s, email=%s\n",
         config.version, config.name, config.email);
+    free((void *) config.name);
+    free((void *) config.email);
```
benhoyt: Couple of nits:

* Can you please put a blank lines before and after the free block, i.e., before line 43 and after line 44, so they're kind of a standalone unit?
* Is the `(void *)` cast to avoid const warnings? If so, can you please format it without additional whitespace for consistency with [this style](https://github.com/benhoyt/inih/blob/535a7f7aef7cf1034a963d07806ea0a2a29c4e90/examples/ini_example.c#L18): `free((void*)config.name)`

girish946: Thanks for the review.

>     * Can you please put a blank lines before and after the free block, i.e., before line 43 and after line 44, so they're kind of a standalone unit?

done 

> 
>     * Is the `(void *)` cast to avoid const warnings? If so, can you please format it without additional whitespace for consistency with [this style](https://github.com/benhoyt/inih/blob/535a7f7aef7cf1034a963d07806ea0a2a29c4e90/examples/ini_example.c#L18): `free((void*)config.name)`

Yes it is for avoiding the `const` warnings.
done.

benhoyt:
Great, thank you - merged!

PR #97, opened 2020-02-22
add meson for system install

stephanlachnit:
I don't know if you are interested considering #59, but I wired up meson support anyway. This is purely optional for users, but a nice feature for distros that want to ship your parser.
The specific reason for this is Feral's gamemode, which ship its own (old) version of inih but rather should use a system installation which is easy to link against.

Installing via meson adds `ini.h` to the include path, adds a shared library `libinih.so` which programs can dynamically link against and also adds a pkg-config entry, so programs can find it.
The nice thing about meson is that you don't have to care about any paths, just change the project version for a new release, and in case there is a breaking API change bump the `soversion`.
In case you don't want to change the build file for every release, I can also remove that feature.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,42 @@
+project('inih',
+    ['c','cpp'],
+    meson_version : '>= 0.46.0',
+    default_options : ['default_library=both'],
+    license : 'BSD-3-Clause',
+    version : '47'
+)
+
+pkg = import('pkgconfig')
+
+#### inih ####
+install_headers('ini.h')
+
+lib_inih = library('inih',
+    ['ini.c', 'ini.h'],
+    install : true,
+    version : meson.project_version(),
+    soversion : '1'
+)
+
+pkg.generate(lib_inih,
+    name : 'inih',
+    description : 'simple .INI file parser',
+    version : meson.project_version()
+)
+
+#### INIReader ####
+install_headers('cpp/INIReader.h')
+
+lib_INIReader = library('INIReader',
+    ['cpp/INIReader.cpp', 'cpp/INIReader.h'],
+    dependencies : declare_dependency(link_with : lib_inih),
+    install : true,
+    version : meson.project_version(),
+    soversion : '1'
+)
+
+pkg.generate(lib_INIReader,
+    name : 'INIReader',
+    description : 'simple .INI file parser for C++',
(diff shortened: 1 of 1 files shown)


stephanlachnit:
Just added some more stuff. It now also installs the C++ variant, and static versions of the libraries for developers. It would also be possible to add the tests to meson (and travis), and installation of the examples.

benhoyt:
Yeah, unfortunately I'd rather not be in the business of maintaining N different build systems for this project. I believe that repo ownership should mean I should know it works (and I'd have to learn Meson or CMake, for example). So I'd rather just leave this kind of thing to individuals using it. Sorry about that, and thanks anyway!

stephanlachnit:
> Yeah, unfortunately I'd rather not be in the business of maintaining N different build systems for this project.

That's not what I mean. One build system is enough, which one really doesn't matter that much. If you prefer make, I could go ahead with that.

> I believe that repo ownership should mean I should know it works (and I'd have to learn Meson or CMake, for example).

I disagree here. Since it's optional and not even a code feature, you don't really need to test it. Distro Maintainers will probably come around with a fix if something shouldn't work (which I hardly doubt given how simple it is to build inih).

> So I'd rather just leave this kind of thing to individuals using it. Sorry about that, and thanks anyway!

If every Distro uses a different build method, people who want to use inih on a Desktop system can't really rely on the system installation in terms of enabled features or pkg-config variables, and have to ship to own version of inih, which needs to be updated manually (like for example Feral's gamemode, it uses a 2 year old version of inih).

Please consider adding one build system for Distro maintainers - it helps users and developers. If you want to understand my build file, you can take a look at the [meson reference manual](https://mesonbuild.com/Reference-manual.html), it's only really using three functions (`library()`, `install_headers()` and `generate()` from the Pkgconfig module) and quite simple, should take less than ten minutes.

If you really don't want to add it, can you at least point in the readme to my fork? I can maintain it out of tree, but it only really makes sense if Distro maintainers and developers know about it.

benhoyt:
Thanks for the thoughtful reasoning -- you've convinced me. :-) Reopening, will merge.

stephanlachnit:
Thanks! If people report any issues on the build system, just mention me and I will look what I can do :+1:

PR #98, opened 2020-02-26
more meson stuff

stephanlachnit:
Sorry that I forgot about this the first time, just two little things:
- I added two variables for dependencies, that makes the integration of inih as a subprojects for other programs easier
- Notes in the readme about meson, just to clarify why there is a build file

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,12 @@ Some differences between inih and Python's [ConfigParser](http://docs.python.org
 
 * Windows/Win32 uses UTF-16 filenames natively, so to handle Unicode paths you need to call `_wfopen()` to open a file and then `ini_parse_file()` to parse it; inih does not include `wchar_t` or Unicode handling.
 
+## Meson notes ##
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -128,6 +128,12 @@ Some differences between inih and Python's [ConfigParser](http://docs.python.org
 
 * Windows/Win32 uses UTF-16 filenames natively, so to handle Unicode paths you need to call `_wfopen()` to open a file and then `ini_parse_file()` to parse it; inih does not include `wchar_t` or Unicode handling.
 
+## Meson notes ##
+
+* The `meson.build` file is intended to build libraries which can be installed on a system. This is not required to use or compile inih.
+* If you want to use inih for program which may be shipped in a distro, consider linking against the shared library. Meson adds entries for pkg-config (`inih` and `INIReader`).
```
benhoyt:
```suggestion
* If you want to use inih for programs which may be shipped in a distro, consider linking against the shared library. Meson adds entries for pkg-config (`inih` and `INIReader`).
```

PR #100, opened 2020-02-28
r48 release

stephanlachnit:
So I send commits to gamemode to check for inih with pkg-config, and everything works out just fine :+1: Will take a look at some more programs to link with pkg-config library if possible.

Now distros need to pack inih (already prepared it for Debian), so that programs can integrate it easy in their builds. The easiest way is to do a new release, it's a bit odd since there is only really b195a45a99a59f126dbbc1819ef4d73487c46f93 compared to r47, but I would appreciate a new release, there are still enough integers left :)

I bumped the copyright (if you rebase maybe don't mark me as author in that commit), removed the makefile for the static library since it's not really necessary (there are theoretically still some use cases, so you may decide if you want to keep the commit or not) and bumped the version in the meson build file. I also changed the `soversion` to 0 since it's the first release with a library build file anyway.


benhoyt:
Thanks! Released: https://github.com/benhoyt/inih/releases/tag/r48

PR #101, opened 2020-03-03
add runtime options

yangfl:
Add support to change options during runtime. This makes shared library much meaningful because if applications can't change the options, they will be forced to build a static copy, which makes shared library useless.

The runtime options will only become available if INI_RUNTIME_OPTIONS defined. Projects which embeds the source files will see no difference.

### DIFF
--- a/ini.c
+++ b/ini.c
@@ -16,17 +16,33 @@ https://github.com/benhoyt/inih
 #endif
 
 #include <stdio.h>
+#include <stdlib.h>
 #include <ctype.h>
 #include <string.h>
 
 #include "ini.h"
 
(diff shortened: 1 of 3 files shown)


benhoyt:
Thanks for your contribution. However, I'd really rather not do this -- it *significantly* changes the code. And may add size and runtime weight, especially for simple compilers used on embedded systems. Inih is very simple and was never designed for use in a shared library, it was designed to use statically, particularly in small constrained systems like embedded microcontrollers.

So I'm not going to merge this. However, feel free to create your own dynamic/shared fork of inih.

PR #103, opened 2020-04-20
Add options for Meson

stephanlachnit:
With this PR I want to add the possibility to _either_ install inih as a shared library _or_ to just build it as a static library. Installing it by default causes some minor inconveniences for some downstream users, so making it optional seems a good idea. I also sneaked in two minor commits.

WiP since I currently only implemented the installation option.

Is it okay to mark this as r49 once implemented? Would get it faster to downstream users.

/cc @kakra, ref FeralInteractive/gamemode#222.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,8 +130,13 @@ Some differences between inih and Python's [ConfigParser](http://docs.python.org
 
 ## Meson notes ##
 
-* The `meson.build` file is intended to build libraries which can be installed on a system. This is not required to use or compile inih.
-* If you want to use inih for programs which may be shipped in a distro, consider linking against the shared library. Meson adds entries for pkg-config (`inih` and `INIReader`).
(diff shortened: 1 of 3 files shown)


kakra:
@stephanlachnit Thanks for the update... But I'll probably arrange with making a separate package for inih since it looks easy enough. The sneaked-in commits still make sense ofc.

stephanlachnit:
Ready to merge from my side.

> @stephanlachnit Thanks for the update... But I'll probably arrange with making a separate package for inih since it looks easy enough. The sneaked-in commits still make sense ofc.

I still think this is a good idea, it makes the Meson setup way more flexible for different use cases (even though the main purpose will never change). Be sure to check out the new options when building the package if this gets merged.

kakra:
I'm quite new to meson but semantically it looks good to me.

PR #106, opened 2020-06-17
Add remove leading quote

nolange:


benhoyt:
Thanks for the contribution. However, `inih` is a very minimalist library, and I'd rather not start supporting quote-parsing options. Quotes aren't supported by Python's `configparser`, which is what inih parsing is based on (they're not supported by many other ini parsers either). If people want this, they can just strip the quotes in the app-specific handler function ... or use a full-blown format like TOML or JSON.

nolange:
Seems a bit arbitrary, but OK.

The change in question is commit 4, so I will open another PR with just this one removed shortly. (contains some improvements Incl. fix for #107)

benhoyt:
Thanks - yep, looking at #107 now.

PR #108, opened 2020-06-19
Improve string operations

nolange:
This series improves the code by removing repeated calls to strlen functions (or hand-rolled equivalents). Instead it used the already known size, an uses a plain memcpy.

### DIFF
--- a/ini.c
+++ b/ini.c
@@ -72,16 +72,12 @@ static char* find_chars_or_comment(const char* s, const char* chars)
     return (char*)s;
 }
 
-/* Similar to strncpy, but ensures dest (size bytes) is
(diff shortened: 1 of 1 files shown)


benhoyt:
Hi @nolange -- thanks for your contributions here, but I'm not going to merge this. Utmost performance isn't a goal for inih, and this change adds a bit of complexity for very little benefit. There's enough going on here that it'd take me more time to review than I have right now (and I'm being a bit ultra-conservative due to messing up the strncpy-to-memcpy fix -- thanks again for that bug report).

Feel free to fork inih if you need various improvements (e.g., refactorings like this one, or quote handling), as I'm probably not going to be merging much except very specific bug-fixes at this point.

PR #109, opened 2020-06-19
meson: remove project version

stephanlachnit:
Isn't really needed and only bloats the commit history (if bumped for every release) for this small lib.


benhoyt:
Looks good to me -- thanks!

PR #112, opened 2020-07-26
Meson tests

Liambeguin:
Mesonbuild comes with a simple test framework. 
Since it's already a build dependency, update test scripts to use it to improve test reports.


benhoyt:
Thanks for the contribution. Personally I'm not a big fan, as I don't know anything about Meson and have never used it -- I would much prefer to write and maintain tests in basic, built-in tools that I understand. @stephanlachnit, as the original author of the [Meson build/install system](https://github.com/benhoyt/inih/pull/97), any thoughts on this?

stephanlachnit:
@benhoyt the nice thing about tests with Meson is the interface. If you run `meson test -C builddir`, it will show you which tests failed (if any), how long each test took, and it logs the test commands and output to a file so you focus on the important stuff.

I don't have the time to take a deep look at this right now, the code looks rather complicated for Meson but that doesn't have to mean anything. I'll try to give it a review the next couple of days.

Since you are the maintainer, I don't think it makes sense to migrate to a test system that you don't understand. However, it should be possible to rewrite this so it works side by side with a script based test system like the current one.
As long as someone (either me or @Liambeguin) takes care that the Meson code is maintained, I don't see anything against it.

.travis.yml
```diff
@@ -1,11 +1,20 @@
...
 # Setting sudo access to false will let Travis CI use containers
 # rather than VMs to run the tests. For more details see:
 # https://docs.travis-ci.com/user/reference/overview/
 sudo: false
 
+before_install:
+  - sudo apt-get install python3-pip ninja-build
+  - pip3 install --upgrade pip
+  - pip3 install --user meson
+  - meson --version
```
stephanlachnit, line 13: Not a big fan of this. Rolling Meson releases make the CI kinda unpredictable. Better to install Meson from Ubuntu, also stuff installed via apt should be done with the apt addon, see https://github.com/stephanlachnit/editorconfig-ctest/blob/d32d7582d551442de77cfd5dd8a6e2f8ad146c57/.travis.yml#L4-L10

.travis.yml
```diff
@@ -1,11 +1,20 @@
+dist: bionic
```
stephanlachnit, line 1: Why not focal? That's the latest LTS.

.travis.yml
```diff
@@ -1,11 +1,20 @@
...
 
+before_install:
+  - sudo apt-get install python3-pip ninja-build
+  - pip3 install --upgrade pip
+  - pip3 install --user meson
+  - meson --version
+
 script:
+  - meson setup build
+  - ninja -C build/
```
stephanlachnit: Please use `builddir` as build directory. See https://github.com/mesonbuild/meson/issues/1560.

I prefer to use Meson to compile:
```sh
meson builddir
meson compile -C builddir
```

.travis.yml
```diff
@@ -14,7 +14,7 @@ before_install:
 
 script:
   - meson setup build
-  - ninja -C build/
```
stephanlachnit: I would still leave the compile command as a single step. It's easier to see whether the compilation or the test failed.

tests/meson.build
```diff
@@ -0,0 +1,86 @@
+# keep these in alphabetical order to not misalign return codes
+inifiles = [
```
stephanlachnit, line 2: Using `files()` is a bit prettier:
```meson
ini_files = files([
    '...',
])
```

Liambeguin:
Hi,

Thanks to both of you for taking the time to look at this.

@benhoyt I understand your reservation.
As @stephanlachnit pointed it out, the reason why I started this patch was to try and make the test results easier to read.
Ideally, Meson would only be used to feed data to a simple test application (like for `tests/unittest_string.c`).
I don't really like how I check the return codes with `tests/unittests.c` but I didn't want to change too much of the tests at once.

If you'd rather, I could work on finding a way to get both test systems to work side by side and I'd be happy to help with the maintenance/review of the Meson code.

.travis.yml
Liambeguin: I'll update to use `focal` instead, I didn't see it in the documentation.
That should allow to no longer have to rely on `pip` too.

.travis.yml
Liambeguin: thanks for the example, will update

.travis.yml
Liambeguin: I switched to `builddir`.
I agree with you that `meson compile` is better but it was only added in `Meson 0.54.0` which is too recent for `focal`.
I'd leave `ninja -C builddir` for now and switch when it's available in a distribution package.

benhoyt:
Thanks both of you for your time on this. I've thought about this further, and for better or worse I'm not really interested in spending time to learn Meson (as the maintainer I want to be able to understand what's going on in the tests). So I'd rather the current "snapshot tests" weren't removed. However, I'd be willing to consider something like this if 1) the Meson test step could *run* the existing tests, or 2) if new Meson tests could live alongside the existing ones (though that would have the disadvantage of repeating the tests). Is option 1 possible?

benhoyt:
To add to that: the reason I'm not particularly concerned about having nice test output etc is because inih is quite stable and changes very rarely. So having pretty or detailed test output is not really what I'm optimizing for.

Liambeguin:
Hi @benhoyt, 

I understand that the project doesn't really need nice test outputs. Also, given the fact that you're not really interested in learning Meson, this might not be so interesting. 
Thanks for your time,

benhoyt:
Thanks for understanding, and sorry about that.

PR #113, opened 2020-08-02
Meson: fix the integer conversions

xx55tt:
Meson doesn't have a `str` function, so it's not possible to change the integer options.
`subprojects/inih/meson.build:50:8: ERROR: Unknown function "str".`


benhoyt:
This looks right to me from a quick glance at the Meson syntax page. @stephanlachnit ?

PR #114, opened 2020-08-04
allow ']' in section name

Liambeguin:
In some cases, a section name might contain a `]`.
Make sure it gets included in the section name by using the last non-comment `]` as end-of-section.

For my specific use-case, the section names contain regular expressions which might have `[` and `]`.

### DIFF
--- a/ini.c
+++ b/ini.c
@@ -71,6 +71,35 @@ static char* find_chars_or_comment(const char* s, const char* chars)
     return (char*)s;
 }
 
+/* Return a pointer to the last non-comment ']' or the first inline comment in
+   a given string, or pointer to NUL at end of string if neither found.
+   Inline comment must be prefixed by a whitespace character to register as a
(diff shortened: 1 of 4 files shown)


benhoyt:
I'd rather inih didn't accept this -- it adds complexity, and most INI parsers don't accept this (including Python's ConfigParser, which `inih` is loosely based off):

```
$ python3
>>> import configparser as cp
>>> p=cp.ConfigParser()
>>> p.read_string('[sec[inner]outer]\nk2=v2')
>>> dict(p)
{'DEFAULT': <Section: DEFAULT>, 'sec[inner': <Section: sec[inner>}
```

In general I think section names and key names (not values) should be simple alphanumeric identifiers, and a regex should be specified as a field value. For example:

```
[section]
pattern = [a-z0-9]+
thing = foo
bar = 1234
```

PR #117, opened 2020-09-27
enable lineno by default

stephanlachnit:
Pretty straight forward, doesn't change any functionality.

This is mainly useful when building the shared lib, because passing a build option can't change the header file. This is also the behavior of the shared lib in Debian ([patch](https://salsa.debian.org/yangfl-guest/inih/-/blob/33817ff191963d08d8de4a06534471cb4333fb9b/debian/patches/0005-Force-INI_HANDLER_LINENO.patch)).

The only thing I'm not completely happy with are the tests, as the script now looks a bit more ugly. If it bothers you, I'm sure I can come up with a cleaner solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ You can control various aspects of inih using preprocessor defines:
 ### Parsing options ###
 
   * **Stop on first error:** By default, inih keeps parsing the rest of the file after an error. To stop parsing on the first error, add `-DINI_STOP_ON_FIRST_ERROR=1`.
-  * **Report line numbers:** By default, the `ini_handler` callback doesn't receive the line number as a parameter. If you need that, add `-DINI_HANDLER_LINENO=1`.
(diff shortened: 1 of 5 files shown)


benhoyt:
I'm not sure I want this on by default (in simple use cases it's not needed). Also, doesn't the change to `ini.h` mean that this will break everyone's code (who include ini.h directly), as the function signature is different by default?

stephanlachnit:
The "not needed in simple cases" argument is a valid point, but it could also be applied to inline comments or BOM sequences. I just think that in embedded use cases, custom compile commands will be set anyway so it doesn't really matter what the default value is.

It shouldn't break existing code though. The only that happens is that when drop-in replacing the library (static or shared doesn't matter), it breaks already compiled code due to the signature change.

However, a simple recompilation should fix that, usually distros have tools to handle such cases. Anyway, afaik the only user of the shared lib right now is gamemode so it's not a big deal, I send in some patches for imv but they haven't been merged yet.

benhoyt:
> it breaks already compiled code due to the signature change.

Yeah, that's what I'd rather not do. I'm going to close this -- turning it on is only a `#define` away.

Issue #123, opened 2020-12-12
pkg-config files have version `undefined`


maxice8:
This breaks Alpine Linux which rely on `Version: ` to be something usable to automatically generate dependencies for `-dev` packages.

```pkgconfig
prefix=/usr
libdir=${prefix}/lib
includedir=${prefix}/include

Name: inih
Description: simple .INI file parser
Version: undefined
Libs: -L${libdir} -linih
Cflags: -I${includedir}
```

benhoyt:
Unfortunately I'm not familiar with pkg-config (or the meson.build files that I guess generates this). @stephanlachnit, any idea of the best approach here?

stephanlachnit:
This is indeed correct, it used to have a version but this requires an update of `meson.build` on every update, which didn't always happen resulting in wrong versions.

I would just add it downstream (by patching Meson with three lines) if it's important.

maxice8:
Understood, added downstream.

PR #125, opened 2021-02-05
enable distro settings by default

stephanlachnit:
inih as a shared library becomes increasingly popular and is even used by xfsprogs now. Because of that, I think it's a good idea to enable these settings in Meson by default. Since Meson supports individual settings per subproject, this won't break any use case.

I also added a small note so that issues like #123 don't appear again.


README.md
```diff
@@ -132,13 +132,14 @@ Some differences between inih and Python's [ConfigParser](http://docs.python.org
 ## Meson notes ##
 
 * The `meson.build` file is not required to use or compile inih, its main purpose is for distributions.
-* By default Meson only creates a static library for inih, but Meson can be used to configure this behavior:
-* with `-Ddefault_library=shared` a shared library is build.
-* with `-Ddistro_install=true` the library will be installed with the header and a pkg-config entry, you may want to set `-Ddefault_library=shared` when using this.
-* with `-Dwith_INIReader` you can build (and install if selected) the C++ library.
-* all compile-time options are implemented in Meson as well, you can take a look at [meson_options.txt](https://github.com/benhoyt/inih/blob/master/meson_options.txt) for their definition. These won't work if `distro_install` is set to `true`.
+* By default Meson is set up for distro installation, but this behavior can be configured for embedded use cases:
+* with `-Ddefault_library=static` static libraries are built.
```
benhoyt, line 136: Thanks for this. One little tweak: can you please indent the three "with" bullets -- they should be sub-points of the "By default" bullet, right?

benhoyt (approved):

README.md
stephanlachnit: Done 👍🏼

PR #126, opened 2021-02-07
meson: add static compile args to inih_dep

stephanlachnit:
Just noticed a small bug, when setting static compile commands in Meson, they haven't been applied to objects that depend on it. Since this might created some bugs, it would be cool if you could tag this a new release (even though no code changed since r52).


benhoyt:
Thanks! Tagged r53: https://github.com/benhoyt/inih/releases/tag/r53

PR #134, opened 2022-03-22
Add visibility symbols

Tachi107:
They are required to properly build DLLs on Windows, and improve the quality of shared objects on Linux. See https://gcc.gnu.org/wiki/Visibility for details.

This issue was first discovered here: https://github.com/mesonbuild/wrapdb/pull/340#issuecomment-1075102565

CC @stephanlachnit and @eli-schwartz

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -6,6 +6,7 @@ project('inih',
 #### options ####
 arg_static = []
 distro_install = get_option('distro_install')
+extra_args = []
 
 if distro_install
     pkg = import('pkgconfig')
@@ -55,15 +56,26 @@ else
     endif
 endif
 
+if host_machine.system() == 'windows'
+    lib = get_option('default_library')
+    if lib == 'both'
+        error('default_library=both is not supported on Windows')
+    elif lib == 'shared'
+        extra_args += '-DINI_SHARED_LIB'
+        add_project_arguments('-DINI_SHARED_LIB_BUILDING', language: ['c', 'cpp'])
+    endif
+endif
+
 #### inih ####
 inc_inih = include_directories('.')
 
 lib_inih = library('inih',
     ['ini.c'],
     include_directories : inc_inih,
-    c_args : arg_static,
+    c_args : [arg_static, extra_args],
     install : distro_install,
-    soversion : '0'
+    soversion : '0',
+    gnu_symbol_visibility: 'hidden'
 )
 
 if distro_install
@@ -72,12 +84,13 @@ if distro_install
     pkg.generate(lib_inih,
         name : 'inih',
         description : 'simple .INI file parser',
+        extra_cflags : extra_args,
     )
 endif
 
 inih_dep = declare_dependency(
     link_with : lib_inih,
-    compile_args : arg_static,
+    compile_args : arg_static + extra_args,
     include_directories : inc_inih
 )
 
@@ -88,10 +101,12 @@ if get_option('with_INIReader')
 
     lib_INIReader = library('INIReader',
         ['cpp/INIReader.cpp'],
+        cpp_args : extra_args,
         include_directories : inc_INIReader,
         dependencies : inih_dep,
         install : distro_install,
-        soversion : '0'
+        soversion : '0',
+        gnu_symbol_visibility: 'hidden'
     )
 
     if distro_install
@@ -100,11 +115,13 @@ if get_option('with_INIReader')
         pkg.generate(lib_INIReader,
             name : 'INIReader',
             description : 'simple .INI file parser for C++',
+            extra_cflags : extra_args,
         )
     endif
 
     INIReader_dep = declare_dependency(
         link_with : lib_INIReader,
-        include_directories : inc_INIReader
+        include_directories : inc_INIReader,
+        compile_args : extra_args
     )
 endif
--- a/cpp/INIReader.h
+++ b/cpp/INIReader.h
@@ -15,6 +15,27 @@
 #include <map>
 #include <string>
 
+// Visibility symbols, required for Windows DLLs
+#ifndef INI_API
+#if defined _WIN32 || defined __CYGWIN__
+#	ifdef INI_SHARED_LIB
+#		ifdef INI_SHARED_LIB_BUILDING
+#			define INI_API __declspec(dllexport)
+#		else
+#			define INI_API __declspec(dllimport)
+#		endif
+#	else
+#		define INI_API
+#	endif
+#else
+#	if defined(__GNUC__) && __GNUC__ >= 4
+#		define INI_API __attribute__ ((visibility ("default")))
+#	else
+#		define INI_API
+#	endif
+#endif
+#endif
+
 // Read an INI file into easy-to-access name/value pairs. (Note that I've gone
 // for simplicity here rather than speed, but it should be pretty decent.)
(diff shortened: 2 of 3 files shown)


meson.build
```diff
@@ -55,15 +56,20 @@ else
     endif
 endif
 
+if get_option('default_library') != 'static'
+    extra_args += '-DINI_SHARED_LIB'
+endif
```
eli-schwartz, line 61: This is only necessary on Windows, and I *think* only with MSVC?, and in that case you need a bit more complicated of a check:
- if default_library is shared, pass that arg
- if default_library is both, error out with the following message: `error('default_library=both is not supported with MSVC')`
- if default_library is static, skip

meson.build
```diff
@@ -55,15 +56,20 @@ else
...
+
 #### inih ####
 inc_inih = include_directories('.')
 
 lib_inih = library('inih',
     ['ini.c'],
     include_directories : inc_inih,
-    c_args : arg_static,
+    c_args : [arg_static, extra_args, '-DINI_SHARED_LIB_BUILDING'],
```
eli-schwartz, line 69: This option is not *needed* except on Windows / MSVC, anyway you can "probably" use add_project_arguments() for this.

meson.build
```diff
@@ -72,12 +78,13 @@ if distro_install
     pkg.generate(lib_inih,
         name : 'inih',
         description : 'simple .INI file parser',
+        extra_cflags : extra_args,
```
eli-schwartz, line 81: By only adding that value on MSVC, not on Linux, you can reduce unneeded arguments included in pkg-config files on Linux.

eli-schwartz (changes requested):

meson.build
eli-schwartz: The problem here is that with default_library=both you build both static and shared, but there is no option to define *different* defines for each. And my understanding is MSVC is so... interestingly designed... that it downright breaks your library if you add a sharedlib declspec when building statically. This is not a problem with GCC or clang.

Tachi107: > This is only necessary on Windows, and I think only with MSVC?

This is necessary on Windows, but it's nice to have on other platforms too, as it makes .so files cleaner and lighter.

Also, I'm not sure that erroring out is necessary, as you can actually build `both` libraries without issues even on Windows.

I check for `!= 'static'` instead of `== 'shared'` because while you can compile the static library when `-DINI_SHARED_LIB` is defined, you can't compile the shared one when the macro is absent.

eli-schwartz: > This is necessary on Windows, but it's nice to have on other platforms too, as it makes .so files cleaner and lighter.

No, I mean the -D define is unnecessary on other platforms, because on `__GNUC__` (gcc/clang) you just unconditionally --without checking for `INI_SHARED_LIB` -- define an `__attribute__ ((visibility ("default")))`.

And thus, even without the define, your .so files are always cleaner and lighter. Because sane compiler.

> Edit2: so, as a `BothLibraries` object in Meson behaves like a `SharedLibrary` when used, defining `INI_SHARED_LIB` when the lib type is `both` or `shared` seems reasonable to me.

The issue is that it will also *install* a `StaticLibrary`, and then attempting to use that from other build systems which are configured to search for and prefer static libraries, may produce broken results.

Given that the default is shared, and defining `both` is something you'd need to opt into, the default behavior is fine anyway.

benhoyt:
Thanks for the submission. Hmm, I don't mind messy build files, but I'm not a fan of polluting the `ini.h` and `INIReader.h` source code with this stuff. I will merge if we really need to, but is there no other way to make this work for Windows DLLs, e.g., with a command line parameter when linking?

Interested in others' thoughts too.

eli-schwartz:
The thing is that some of this "mess" is needed anyway. You want to *unconditionally* define it to `__attribute__ ((visibility ("default")))` on GCC/clang or anything else that supports the visibility attribute, because that leads to better GCC/clang optimizations and a leaner symbol table.

It's a one-line define on GCC/clang, plus adding the macro in front of each public symbol.

The problem is of course the MSVC compiler, which has... "interesting" design and requires a complex maze of conditions to determine whether or not to pass MSVC's competing attribute, declspec. And also, which *value* to use for the MSVC attribute.


MSVC also has the concept of .def files, which can be passed in meson using `vs_module_defs: 'inih.def'` and you get to manually maintain that file which has a list of public symbols.

Of course, you want to use GCC symbol visibility regardless of what you do with MSVC.

(Technically, GCC can do like the .def file too. GCC calls it an [ld version script](https://sourceware.org/binutils/docs-2.37/ld/VERSION.html), and it uses a totally different format, but again, it's a list of public symbols marked as global, and a wildcard on all remaining local symbols. Use of an ld version script does NOT enable GCC to perform compiler optimizations on symbols that aren't public, because it doesn't know until link time that they aren't public. At link time, the private symbols still exist, they just get hidden.)

meson.build
Tachi107: > No, I mean the -D define is unnecessary on other platforms, because on `__GNUC__` (gcc/clang) you just unconditionally --without checking for `INI_SHARED_LIB` -- define an `__attribute__ ((visibility ("default")))`.

Oh, you're right, I copied the various `#if`s from another project of mine and didn't noticed that this was the case 😅️

> The issue is that it will also _install_ a `StaticLibrary`, and then attempting to use that from other build systems which are configured to search for and prefer static libraries, may produce broken results.

Yep, calling `error()` on Windows might be the only solution.

> (I am no expert on Windows or MSVC. But I've been told that it "does" produce broken results -- you end up building both a dll and a static library, and the dll works, and the static library doesn't.)

Awesome.

Tachi107:
> Thanks for the submission. Hmm, I don't mind messy build files, but I'm not a fan of polluting the `ini.h` and `INIReader.h` source code with this stuff. I will merge if we really need to, but is there no other way to make this work for Windows DLLs, e.g., with a command line parameter when linking?

I know, it sucks, but the other options are even worse in my opinion (as Eli says, you'd have to maintain some files that would need to be (manually?) updated every time you modify your symbols). I don't know exactly how painful they are to adopt because I've never bothered to look into the alternatives, but having that set of `#ifdef`s and the various `NAME_API` is done in a lot of projects

meson.build
Tachi107: By the way I noticed that you can pass `[arg_static, extra_args]` to `library.c_args` but not `declare_dependency.compile_args` (it works with `arg_static + extra_args` though) - _"List item must be one of <class 'str'>, not <class 'list'>"_

eli-schwartz: Some places do implicit flattening at time of use, some do not. Personally I find implicit flattening makes me nervous.

benhoyt:
Okay, thanks. I'll go ahead and merge this, given that it seems there aren't good workarounds. @stephanlachnit Could I please get your thumbs-up first?

benhoyt:
Thanks for the contribution, @Tachi107!

benhoyt:
@stephanlachnit r54 tagged and released: https://github.com/benhoyt/inih/releases/tag/r54

eli-schwartz:
Thanks, inih is now updated on the WrapDB with passing CI for Windows/macOS. :)

Tachi107:
Thanks for merging!

PR #135, opened 2022-03-24
meson.build: define version

thesamesam:
Otherwise, the installed .pc file contains "Version: undefined".

Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,7 @@
 project('inih',
     ['c'],
     license : 'BSD-3-Clause',
+    version : '54',
 )
 
 #### options ####


benhoyt:
Thanks for the submission. Care to comment @stephanlachnit and @Tachi107?

eli-schwartz:
This seems correct. From the Meson documentation on pkgconfig.generate (and similar wording for declare_dependency):
>  `version` a string describing the version of this library, used to set the `Version:` field. (since 0.46.0) Defaults to the project version if unspecified.

When building inih from git master:
```
$ meson setup builddir
The Meson build system
Version: 0.62.0
Source dir: /tmp/inih
Build dir: /tmp/inih/builddir
Build type: native build
Project name: inih
Project version: undefined      # <<<<<<<<<<<<
C++ compiler for the host machine: ccache c++ (gcc 11.1.0 "c++ (GCC) 11.1.0")
C++ linker for the host machine: c++ ld.bfd 2.36.1
Build targets in project: 2
```

When building it with this PR:
```
$ meson setup builddir
The Meson build system
Version: 0.62.0
Source dir: /tmp/inih
Build dir: /tmp/inih/builddir
Build type: native build
Project name: inih
Project version: 54             # <<<<<<<<<<<<<<<<
C++ compiler for the host machine: ccache c++ (gcc 11.1.0 "c++ (GCC) 11.1.0")
C++ linker for the host machine: c++ ld.bfd 2.36.1
Build targets in project: 2
```

Defining the version means that it is possible to trivially require a specific version as a dependency using pkg-config or a Meson subproject.

Tachi107:
This seems appropriate, thanks :)

stephanlachnit:
See #123.

Tachi107:
Yeah it needs an update on every release but it seems that they don't occur that often, so it shouldn't be a huge maintenance burden. Of course @benhoyt is the one who should ultimately decide whether or not he wants to update it on every release

eli-schwartz:
Perhaps a release script would help? Also, recent versions of Meson can read the version from a `files('VERSION.txt')` which may be a lot easier to automate bumping the version in.

And also in the long term I someday plan to add support for setuptools_scm like automatic git-based versioning.

benhoyt:
@thesamesam Do the versions of Meson that matter support this `files()` construct? If so, I'd prefer to have a separate `VERSION.txt` file at the top level of the repo, as @eli-schwartz suggested above.

eli-schwartz:
https://mesonbuild.com/Reference-manual_functions.html#project

You would need to set the `project(..., meson_version: '>=0.57.0')`.

It was released on Feb 14, 2021.

Debian stable has 0.56 but stable backports has 0.61.2 and you can pip install it either way on any distro that isn't stuck with python 3.5 that is practically ancient on top of being end of life.

benhoyt:
Thanks @eli-schwartz. Hmmm, seems maybe a bit premature to me in that case, given key distros don't even have that version. I believe `meson.build` will be the only place the version is mentioned, so let's just use that for now. I'll merge this.

benhoyt:
I've bumped up to 55 in a subsequent commit and release r55: https://github.com/benhoyt/inih/releases/tag/r55

eli-schwartz:
> Thanks @eli-schwartz. Hmmm, seems maybe a bit premature to me in that case, given key distros don't even have that version.

Yeah, many projects do consider supporting at least the version of Meson in the latest stable/LTS of popular distros to be a reasonable concern, hence why I mentioned this doesn't fulfill that criteria.

There are also big projects with people working to port to Meson, who consider 0.54 to be the cutoff because they need to support versions of Red Hat which are still considered supported, but which only have python 3.5. That's a bit extreme IMO as far as support matrixes go, but then, postgres does have some major corporate commitments to make, so it is what it is. Hence I mention that python bound too.

It's not so bad to have it in meson.build itself, anyway. And you can still script it to bump the version, by running meson itself.

https://mesonbuild.com/Rewriter.html#setting-the-project-version

(This is a bit more reliable than sed.)

thesamesam:
Thanks all! I think we can revisit it if desirable in a few.. uh, years 😄.

Appreciate @eli-schwartz's input too as The Meson Guru.

PR #136, opened 2022-04-13
Added tipi.build support and a simple usage example

pysco68:
Hello, Yannic from [tipi.build](https://tipi.build) here,

I just added `tipi` support to your project and was wondering if you'd like to merge so everyone gets a new way of using your project.

I tested the project (both for project dev and as dep) for Linux, MacOS and Windows (both clang and MSVC builds btw) and it seems to work nicely.

If there's any issue or if you have any question with this I'll be happy to sort it out

Best from Zürich,
Yannic


benhoyt:
Hi @pysco68, thanks for the contribution. However, I'd prefer not to add multiple build systems to the inih repository. I (somewhat reluctantly) merged the meson.build configuration, but even then I don't really know how it works. I'm okay keeping meson.build as it's in already, but I'd rather not add another. I suggest that you fork and add support for your build system in your own fork. Sorry about that!

pysco68:
hi @benhoyt I get your point about the build system obscurity and complexity. 

Just on a second thought; would you mind accepting a PR with just the changes to the readme (minus the ref to the tipi example)? That way at least the usage with tipi would be documented "close to" the original repo. The empty `.tipi/deps` file is really just there for us to make compatible repositories discoverable.

For your ref: what tipi basically does is to automatically generate build scripts by "figuring out" which parts fit together. To use a dependency you bascially point the tool at a github repo and it does the magic mostly on its own. If you find that cool I'd be happy to see you try or to give you a short demo if you want.

benhoyt:
@pysco68 Done in https://github.com/benhoyt/inih/commit/4bd3261ea422a99aa764e63820e16d19cdad33dd

eli-schwartz:
I've never heard of tipi.build before, but your link is to a marketing website full of buzzwords like "sign up", "pricing", and "smart remote compiler".

Is this usable as a regular build system? Is this `.tipi/deps` file useful for something other than some commercial service offering?

pysco68:
@benhoyt thank you! Much appreciate the merge 👍 

@eli-schwartz we want to defragment the ecosystem with its dozen or so (at least) "widespread" build systems by removing the need to manually write build scripts (that's a part of the "smart" in the marketing wording you're mentioning). We do so by inferring the build layout and generating (currently only CMake) build scripts for any given project. In that regard we're a "meta-meta-buildsystem", but that's really only a small part of what we're addressing.

Regarding the `.tipi/deps` file, it's basically a description of a project's dependencies (and dependency tree if the libraries that are being used don't have their own `.tipi/deps` yet). Additionally it may contain some hints for the build algorithm when the algorithms isn't smart enough (yet - again ;)) to figure it out on its own. 

In the example of `inih` the file is empty (and thus optional) because there's no dependency that tipi needs to load to build the library. We prefer for compatible repositories (even in simple cases like this one) to have their own `.tipi/deps` because it makes it discoverable for tipi users (and soon for us as we are preparing a searchable index at the moment).

Here's another example of a deps file which contains a more complete suite of things: https://github.com/tipi-deps/tomlplusplus/blob/master/.tipi/deps

Regarding the commercial offering: yes, tipi.build is backed by a company that hopes to make some money out of it. All the core features like the dependency management, tool distribution, source scan etc are and will stay free to use. We do provide paid cloud ressources (stable ad-hoc environment for Linux, Windows and MacOS builds to make the cross-platform dev life easier), and some premium features (build parallelity optimization, build and binary optimization, private build cache, code analytics) that are more enterprise-y.

eli-schwartz:
None of this really answers my question, which was a pretty simple one:

If as a user I read the inih README and see tipi.build mentioned, and say "oh hmm, a new build system, which inih suggests I can use here. I wonder how it works and if it can help me / is easier to use"...

... how do I in fact use it? Is there a tool I download and run locally to build my local code using your algorithm? The link in the README implies that no, I need to create an account on your cloud and submit a build to your cloud.

By the way: you're not the first people to think about build systems detecting how the code should be built without writing Makefiles or suchlike to do it. https://github.com/capnproto/ekam

(ekam introduces itself by explaining how you download and run it, and has a nice layperson's description of how that works.)

PR #138, opened 2022-07-06
Add More Fuzzer Support

0x34d:


benhoyt:
Hi @0x34d, I'm not opposed to adding this, but could you please add a bit more detail in the description? Why we should add this, what your motivation was, what distinguishes it from the existing fuzzing code, have you found any bugs in inih through clang's fuzzer, and so on.

Issue #133, opened 2021-09-02
Do you plan support buid with cmake


aengusjiang:
Cmake is a good across building tool. Do you plan support buid with cmake？

benhoyt:
@aengusjiang Sorry, I don't plan to do this. I personally don't want to keep up with N different build systems. There's already a Meson config file for Meson. See further discussion at https://github.com/benhoyt/inih/pull/105 and https://github.com/benhoyt/inih/pull/97

Ryanf55:
How do you recommend `CMake` users consume this library who want to use the package provided by their distribution? 

From reading some other docs, it appears we should use the pkgconfig file, which appears to be shipped with most distributions.

https://cmake.org/cmake/help/latest/module/FindPkgConfig.html

neheb:
Yes, pkgconfig is the way

Btw if you want to add cmake support, meson supports these two: https://mesonbuild.com/CMake-module.html#cmakewrite_basic_package_version_file

simple example:

```
--- a/meson.build
+++ b/meson.build
@@ -10,6 +10,7 @@ distro_install = get_option('distro_install')
 extra_args = []
 
 if distro_install
+    cmake = import('cmake')
     pkg = import('pkgconfig')
 else
     if not get_option('multi-line_entries')
@@ -82,6 +83,11 @@ lib_inih = library('inih',
 if distro_install
     install_headers('ini.h')
 
+    cmake.write_basic_package_version_file(
+      name : 'inih',
+      version : meson.project_version(),
+    )
+
     pkg.generate(lib_inih,
         name : 'inih',
         description : 'simple .INI file parser',
```

Would simplify inclusion into CMake/Conan but currently the only option is pkgconfig.

Issue #71, opened 2018-10-03
CMake and pkgconfig support


otavio:
We have been using this library for a while (since 2011) and we made a few changes on it. We'd like to contribute them back if you think it makes sense.

- CMake
- pkgconfig

Please let me know if a PR is desired. I can rebase our changes and prepare it.

benhoyt:
Hi @otavio, thanks for that. I'm uncertain about including make/build stuff in such a tiny project. But I don't know much about CMake or pkgconfig, so I'm open to hearing more or seeing your PR.

otavio:
It is a widely used build system and it integrates way with cross-building environments. We did change the library to build static libraries for C and C++ and the pkgconfig allows for easy integration on existing projects.

A snapshot of what we use is here: https://github.com/OSSystems/inih

I need to rework this code and prepare a PR if you think this is worth.

benhoyt:
Hi @otavio: I think I'd rather not include these. Not because they're not useful or a bad idea, but because then I'd have to maintain files/systems I know nothing about, and I don't have much bandwidth for `inih` at this point. Sorry!

PR #141, opened 2022-07-20
add autotools build and GitHub release workflow

bluesquall:
Here's a set of changes that will support the autotools build of a shared & static libraries, and a GitHub release workflow to put the `distcheck`'ed tarball on the release page.

I need these things for a build on a legacy system, and figured they may be useful to others. I could also draft a CMake build, if you're interested.

### DIFF
--- a/.build-aux/git-version-gen
+++ b/.build-aux/git-version-gen
@@ -1,0 +1,1 @@
+printf %s "$(git --no-pager describe --abbrev=8 --tags --always --dirty)"
--- a/.github/workflows/build-if-tag-release.yaml
+++ b/.github/workflows/build-if-tag-release.yaml
@@ -1,0 +1,23 @@
+name: build (and if tag, release)
+on: push
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: autoconf
+        run: autoreconf -fi
+      - name: configure
+        run: ./configure
+      - name: build
+        run: make
+      - name: check
+        run: make check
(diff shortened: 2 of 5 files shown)


benhoyt:
Hi @bluesquall, thanks for your contribution. Unfortunately I've decided I'd rather not keep up with and maintain N build systems, and we've already got support for the fairly well-used Meson build system, so I'm not planning to add others to the main `inih` repo. Feel free to add it in your own fork or repo, though!

See previous similar discussions at https://github.com/benhoyt/inih/pull/105, https://github.com/benhoyt/inih/issues/71, and https://github.com/benhoyt/inih/issues/133.

eli-schwartz:
> I need these things for a build on a legacy system

I assume this means some system which for some reason, python is not ported to it?

Although Meson itself is written in and requires python, there's a c99 port at https://sr.ht/~lattis/muon which can build inih. The muon port is tested on Linux and netbsd, but I don't know what system you're using. It should hopefully work okay on any POSIX-ish system, and at the very least it will be far easier to patch for missing support than it would be to patch both Meson and the entirety of python.

bluesquall:
It is a system that does not have python (at least not usable). I am looking forward to the time when I don’t need to support it anymore, but for now I take what I have and do what I can.

bluesquall:
Fair enough.

I’ll keep an active fork for this platform, and try out meson/muon when I’m on more modern systems.

bluesquall:
If you would be interested in the GitHub actions release workflow using meson instead of autotools, I'd be happy to draft that in a different PR. I totally understand if that is more to maintain, too.

PR #143, opened 2022-08-08
remove quotes from parsed values

angeloc:
Removing quotes while parsin if the INI_NO_STRING_QUOTE is enabled.

Signed-off-by: Angelo Compagnucci <angelo@amarulasolutions.com>

### DIFF
--- a/ini.c
+++ b/ini.c
@@ -48,7 +48,11 @@ typedef struct {
 static char* rstrip(char* s)
 {
     char* p = s + strlen(s);
+#if INI_NO_STRING_QUOTE
+    while (p > s && (isspace((unsigned char)(*--p)) || (unsigned char)(*p) == '"'))
+#else
     while (p > s && isspace((unsigned char)(*--p)))
+#endif
         *p = '\0';
     return s;
 }
@@ -56,7 +60,11 @@ static char* rstrip(char* s)
 /* Return pointer to first non-whitespace char in given string. */
 static char* lskip(const char* s)
 {
+#if INI_NO_STRING_QUOTE
+    while (*s && (isspace((unsigned char)(*s)) || (unsigned char)(*s) == '"'))
+#else
     while (*s && isspace((unsigned char)(*s)))
+#endif
         s++;
     return (char*)s;
 }
--- a/ini.h
+++ b/ini.h
@@ -170,6 +170,13 @@ INI_API int ini_parse_string(const char* string, ini_handler handler, void* user
 #define INI_CUSTOM_ALLOCATOR 0
 #endif
 
+/* Nonzero to remove string quoting from settings. Settings could be saved with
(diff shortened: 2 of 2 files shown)


benhoyt:
Hi @angeloc, thanks for the submission. However, I'd rather not include this change. I want to keep inih fairly minimalist, and it's easy to do this quote removal in the handler if needed.

Also, in my opinion this feature isn't particularly useful without unescaping, for example `sentence = "The \"quick\" brown fox."`. But once again, it would be easy to do with quote removal and unescaping with an `unescape()` function called on the value in the handler.

See a similar feature proposed (`INI_REMOVE_LEADING_QUOTE`) in https://github.com/benhoyt/inih/pull/106.

angeloc:
@benhoyt Not much happy with the decision, honestly, but it's your code after all!
Anyway, just for the sake of discussion, I don't think this proposal has anything to do with escaping.
My problem is that I have a QT software which produces quoted values in INI file and I don't want quotes in the C software consuming it. Quoted values are permitted from the standard: I can understand python confgparser doesn't implement it, but removing quotes from a string in python is ridiculously easy.
Removing quotes is easy in the library and incredibly useful for the user of the library. Implementing another external function only to have quotes removed is overkill and not that easy. You probably end up needing to import another string c library to do some serious (and secure) manipulation.
Anyway, I'll stick to my fork for my project.
Thanks anyway!

benhoyt:
I don't really understand why this needs to be in inih itself, though. To reproduce the functionality in your PR, it seems to me just as easy (and a bit more explicit) to add a call to a simple `strip_quotes()` function, something like so (taken from the example on in the inih README). This is untested, but should be similar to what you did in the PR:

```c
static int handler(void* user, const char* section, const char* name,
                   const char* value)
{
    char* stripped_value = strip_quotes((char*)value); // add this line to your handler function
    // handle section, name, stripped_value
    return 1;
}

// new function: strip quotes from left and right
char* strip_quotes(char* s) {
    char* p = s + strlen(s);
    while (p > s && unsigned char)(*p) == '"') // trim right
        *p = '\0';
    while (*s && (unsigned char)(*s) == '"') // trim left
        s++;
    return s
}

int main(int argc, char* argv[])
{
    configuration config;

    if (ini_parse("test.ini", handler, &config) < 0) {
        printf("Can't load 'test.ini'\n");
        return 1;
    }
    printf("Config loaded from 'test.ini': version=%d, name=%s, email=%s\n",
        config.version, config.name, config.email);
    return 0;
}
```

Using a fork is fine though if you prefer!

angeloc:
Indeed. This implies rescanning all the string not even one but two times, when the library has already done the very same thing ...
The goal of the library should be obtaining the values from the INI, thus removing the quotes is a tiny addiction toward the library easy friendliness imho.

Anyway, after your example, I'm more convinced of my solution ;)

PR #147, opened 2022-11-18
Added a GetUnsigned function for getting unsigned values.

jcormier:
Allow getting unsigned integers without losing precision.

### DIFF
--- a/cpp/INIReader.cpp
+++ b/cpp/INIReader.cpp
@@ -56,6 +56,16 @@ long INIReader::GetInteger(const string& section, const string& name, long defau
     return end > value ? n : default_value;
 }
 
+unsigned long INIReader::GetUnsigned(const string& section, const string& name, unsigned long default_value) const
+{
+    string valstr = Get(section, name, "");
+    const char* value = valstr.c_str();
+    char* end;
+    // This parses "1234" (decimal) and also "0x4D2" (hex)
+    unsigned long n = strtoul(value, &end, 0);
+    return end > value ? n : default_value;
+}
+
 double INIReader::GetReal(const string& section, const string& name, double default_value) const
 {
     string valstr = Get(section, name, "");
--- a/cpp/INIReader.h
+++ b/cpp/INIReader.h
@@ -66,6 +66,11 @@ public:
     // not found or not a valid integer (decimal "1234", "-1234", or hex "0x4d2").
     INI_API long GetInteger(const std::string& section, const std::string& name, long default_value) const;
 
+    // Get an unsigned integer (unsigned long) value from INI file,
+    // returning default_value if not found or not a valid integer 
+    // (decimal "1234", or hex "0x4d2").
+    INI_API unsigned long GetUnsigned(const std::string& section, const std::string& name, unsigned long default_value) const;
+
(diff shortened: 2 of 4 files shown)


cpp/INIReader.cpp
```diff
@@ -56,6 +56,16 @@ long INIReader::GetInteger(const string& section, const string& name, long defau
     return end > value ? n : default_value;
 }
 
+unsigned long INIReader::GetUnsignedInteger(const string& section, const string& name, unsigned long default_value) const
```
benhoyt, line 59: Let's call this `GetUnsigned` for brevity (like with C / C++ itself, the "integer" is kind of obvious).
```suggestion
unsigned long INIReader::GetUnsigned(const string& section, const string& name, unsigned long default_value) const
```

cpp/INIReader.h
```diff
@@ -66,6 +66,10 @@ class INIReader
     // not found or not a valid integer (decimal "1234", "-1234", or hex "0x4d2").
     INI_API long GetInteger(const std::string& section, const std::string& name, long default_value) const;
 
+    // Get an unsigned integer (unsigned long) value from INI file, returning default_value if
+    // not found or not a valid integer (decimal "1234", or hex "0x4d2").
```
benhoyt, line 70: Let's wrap these comment lines at <80 chars, so after "returning" and after "or".

benhoyt (changes requested):
Thanks! I'm good with adding this, though let's call it `GetUnsigned` as mentioned. Also, as you can see by the failed test output, you need to update `examples/cpptest.txt` so the C++ test passes.

benhoyt:
Also, please update the PR title, and add a PR description (ideally with your use case).

jcormier:
Thanks for the feedback. Working on it from my phone 😆. We shall see how that goes. I'll post another update when im done.

jcormier:
Alright changes made. I'd like to squash all these commits together but can't seem to from the phone. So I'll get back to it next week when im back at my computer. Unless your dive with it as is

cpp/INIReader.h
```diff
@@ -66,6 +66,11 @@ class INIReader
     // not found or not a valid integer (decimal "1234", "-1234", or hex "0x4d2").
     INI_API long GetInteger(const std::string& section, const std::string& name, long default_value) const;
 
+    // Get an unsigned integer (unsigned long) value from INI file,
+    // returning default_value if not found or not a valid integer 
+    // (decimal "1234", or hex "0x4d2").
+    INI_API unsigned long GetUnsignedInteger(const std::string& section, const std::string& name, unsigned long default_value) const;
```
benhoyt, line 72:
```suggestion
    INI_API unsigned long GetUnsigned(const std::string& section, const std::string& name, unsigned long default_value) const;
```

benhoyt:
Thanks. Two further changes required: 1) change `GetUnsignedInteger` to `GetUnsigned` in the header file too, and 2) update `examples/cpptest.txt` so the C++ test passes.

jcormier:
Updated and squashed commits

benhoyt:
Thanks!

Issue #145, opened 2022-10-23
Seems to not discard the rest of a long line?


xlucn:
**The problem:**

The issue assumes we are dealing with a long line in an ini file.

In line 136, `reader` (`fgets`) will read `max_line` chars. If using a heap, then in line 139 and forward, the heap will resize and all the rest of the line will be read:

https://github.com/benhoyt/inih/blob/5e1d9e2625842dddb3f9c086a50f22e4f45dfc2b/ini.c#L136-L141

But, in the default case with a stack (fixed size array), `fgets` will read `max_line` and do nothing with the rest of the line, which I think will stay in the input stream. Then, in the next iteration of while on line 136, `fgets` will continue to read that rest of the line, but inih treats those text as the next line. The code will most likely find no `:` or `=` in those text, thus report an error on line #(line number of the long line + 1).

**Discussion:**

So, is this expected behavior, or should inih discard the rest of a lone line? The latter might be something like this:
```c
    while (reader(line, (int)max_line, stream) != NULL) {
#if INI_ALLOW_REALLOC && !INI_USE_STACK
... ...
#else
if (strchr(line, '\n') == NULL)
    while (fgetc(stream) != '\n')
        ;
#endif
```

benhoyt:
Hmm, it's a fair question. However, I'm not sure silently discarding the rest of the line is better than returning an error (and inih will keep parsing on error by default). I guess we could discard the rest of the line to avoid unexpected behaviour, and still return an error?

oliverlew:
> I guess we could discard the rest of the line to avoid unexpected behaviour, and still return an error?

I guess that works, too. Only that "not having read all the texts" is not as severe as a format error or handler-returned error. And currently inih does not provide any other information about the error type (similar to errno).

Anyway, the rest of a long line should be regarded no matter what, in case those texts represent another valid section or key-value entry, thus contaminating the result.

oliverlew:
Could use some test cases like this:
```
a=<197 * '1'>b=222
```
and see if `b=222` is parsed as a new entry. My result with current inih is
```
section: , name: a, value: 11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
section: , name: b, value: 222
```
with this test program
```c
#include <stdio.h>
#include <ini.h>

int handler(void *user, const char *s, const char *n, const char *v)
{
[...]

PR #148, opened 2022-11-23
Fixed handling of long strings when using INI_USE_STACK

tyler92:
Undefined behavior occurred if the incoming .ini file contains a line that is longer than INI_MAX_LINE with the INI_USE_STACK option. In such a case, it is safe to abort parsing even if INI_STOP_ON_FIRST_ERROR is not set.

See https://github.com/benhoyt/inih/issues/145 for the details

Signed-off-by: Mikhail Khachayants <tyler92@inbox.ru>

### DIFF
--- a/ini.c
+++ b/ini.c
@@ -156,6 +156,14 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
 
         lineno++;
 
+#if INI_USE_STACK
+        if (strlen(line) == max_line - 1 && line[max_line - 2] != '\n') {
+            /* Exit even if INI_STOP_ON_FIRST_ERROR is not set */
+            error = lineno;
+            break;
+        }
+#endif
+
         start = line;
 #if INI_ALLOW_BOM
         if (lineno == 1 && (unsigned char)start[0] == 0xEF &&


tyler92:
I'm not sure about INI_STOP_ON_FIRST_ERROR and unit tests. I would appreciate a discussion on this.

ini.c
```diff
@@ -156,6 +156,14 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
 
         lineno++;
 
+#if INI_USE_STACK
```
benhoyt, line 159: I think we should do this check whether or not `INI_USE_STACK` is set, right? Because even in non-stack mode we might have read and realloc'd up to `INI_MAX_LINE` bytes.

ini.c
```diff
@@ -156,6 +156,14 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
...
+        if (strlen(line) == max_line - 1 && line[max_line - 2] != '\n') {
+            /* Exit even if INI_STOP_ON_FIRST_ERROR is not set */
```
benhoyt, line 161: Yeah, I'm not sure about this. If `INI_STOP_ON_FIRST_ERROR` is false (the default), I think we should read and discard the rest of the current line, and then proceed. We could decide to do this as a stop-gap fix, but I'm not sure that's a good idea -- if we fix this I think we should do it properly in a single PR.

You can see how it's breaking the existing max-line tests as it is currently.

benhoyt (commented):
Couple of comments left. Yeah, we'd need to add tests for this. They're "diff tests", meaning they run a program and ensure the output hasn't changed against a file already committed.

See [here](https://github.com/benhoyt/inih/actions/runs/3535066961/jobs/5933912685) for current failures.

tyler92:
OK, thanks for the comments. I'll rework my PR and will come back later.

flysand7:
He never did lmao

Tristan-BS:
@tyler92 you still alive? :0

tyler92:
Alive, but too busy to make changes here. I will close this MR, feel free to open new one with my changes

PR #149, opened 2022-12-02
meson.build: fix start-of-line_comment_prefix variable name

ihilt:
`start-of-line_comment_prefix` is not a valid identifier in Meson.  It looks like this was partially addressed with `sol_comment_prefix` but didn't make it into actual usage.


benhoyt (approved):
Looks right to me, thanks!

PR #151, opened 2023-07-05
Allow INIReader to read 64-bit integers

natcat256:
In some cases, it may be desired to read (unsigned) integer values outside of the 32-bit bounds. In 32-bit code, however, this might _not_ be the case since doing so by default would cost performance. Rather than changing the types delivered by `INIReader::GetInteger` and `INIReader::GetUnsigned`, a good way to accommodate is to add separate methods capable of reading (unsigned) 64-bit integers.


cpp/INIReader.cpp
```diff
@@ -56,6 +56,16 @@ long INIReader::GetInteger(const string& section, const string& name, long defau
     return end > value ? n : default_value;
 }
 
+INI_API long long INIReader::GetBigInteger(const std::string& section, const std::string& name, long long default_value) const
```
benhoyt, line 59: Two things:

1) I think if we keep these type `long long` we should call these `GetLongLongInteger` and `GetLongLongUnsigned`, as they're clearer than "big" (even if a bit, um, long). In my mind "big integer" usually means arbitrarily large like Go's `math/big.Int` types.
2) Is it common in modern C++ to use "long long", or is it more common to use the specific width types like `int64_t` and `uint64_t`? I've been out of the C++ loop for a while.

examples/test.ini
```diff
@@ -8,3 +8,4 @@ name = Bob Smith       ; Spaces around '=' are stripped
 email = bob@smith.com  ; And comments (like this) ignored
 active = true          ; Test a boolean
 pi = 3.14159           ; Test a floating point number
+big = 2147483648       ; Test a big integer
```
benhoyt, line 11: Let's use a larger value for the test integer here. 2147483648 is more than a 32-bit signed long can hold, but fits fine in a 32-bit unsigned long. Maybe 1000000000000?

benhoyt (changes requested):
Thanks for this. I'm happy to add these, however, one small comment about the test value, and a couple of questions about naming and types.

examples/test.ini
natcat256: That was an oversight on my part. Fixed in 0bc41b98f3c28f00c1c264c3520418656e764641

cpp/INIReader.cpp
benhoyt: I've had a quick look around, and it seems `(u)int64_t` is fairly common in C++ codebases. Plus, I think fixed sizes are more sane than "at least this long" sizes. And given that we both prefer it, let's go with that for these new functions, and call them `GetInteger64` and `GetUnsigned64` as you suggest.

I presume you'll have to `#include <cstdint>` to get these?

natcat256: Alright, I updated the code to use those method names and fixed-width types.

These types were already accessible with <map> included, but it's probably best to include <cstdint> anyways, so I have done that.

cpp/INIReader.h
```diff
@@ -66,11 +67,18 @@ class INIReader
...
+    // Get a 64-bit integer (int64_t) value from INI file, returning default_value if
+    // not found or not a valid integer (decimal "1234", "-1234", or hex "0x4d2").
+    INI_API int64_t GetInteger64(const std::string& section, const std::string& name, int64_t default_value) const;
+
+    // Get an unsigned integer (unsigned long) value from INI file, returning default_value if
+    // not found or not a valid unsigned integer (decimal "1234", or hex "0x4d2").
     INI_API unsigned long GetUnsigned(const std::string& section, const std::string& name, unsigned long default_value) const;
 
+    // Get an unsigned 64-bit integer (uint64_t) value from INI file, returning default_value if
+    // not found or not a valid unsigned integer (decimal "1234", "-1234", or hex "0x4d2").
```
benhoyt, line 79: Let's remove the "-1234" example here, as it's incorrect for unsigned.

benhoyt (changes requested):
Looks good now! Requesting one very minor change to remove the "-1234" example from GetUnsigned64.

benhoyt (approved):

benhoyt:
Thanks for your contribution. Just tagged release [r57](https://github.com/benhoyt/inih/releases/tag/r57).

natcat256:
Awesome! Glad I could contribute.

cpp/INIReader.cpp
Clownacy: You do know that `(u)int64_t` is less portable than `(unsigned) long long`? The latter is guaranteed to exist in all conforming C++11-and-later implementations while the former is entirely optional. While this might not be the most relevant on desktop, it could be an issue for embedded uses where toolchains don't implement every optional C++ feature. Using `int64_t` is not simply a matter of coding style: it's for when your code is so frail that it relies on exact size and overflow/underflow behaviour.

benhoyt: @Clownacy Maybe you're right. It looks like they've caused on issue on an old version of GCC: https://github.com/benhoyt/inih/issues/152

That said, I was with you up until your last sentence: it seems to me using fixed-width integers is likely to make code *less* frail, because the width is specified concretely and exactly, rather than "may be this big, may overflow at this point, but may not, you'll never know".

PR #156, opened 2023-07-27
[Fuzzing] fix harness

0x34d:
bug fix: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60878


fuzzing/inihfuzz.c
```diff
@@ -66,12 +66,13 @@ extern int LLVMFuzzerTestOneInput(const char *Data, size_t Size) {
 
     int ret;
     *Prev_section = '\0';
+    static int u = 100;
```
0x34d, line 69: wait i will remove static

PR #157, opened 2023-08-06
Specify C++11 std in meson build

DownerCase:
Compilers differ on what the default standard is and I had to set C++11 standard as part of the [vcpkg port](https://github.com/microsoft/vcpkg/pull/33001) to get it work on MacOS.
Consequence of #142 

I wasn't sure what standard version to put for C so I left it out.

Also you forgot to bump the version number last release.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,8 @@
 project('inih',
     ['c'],
     license : 'BSD-3-Clause',
-    version : '56',
+    version : '57',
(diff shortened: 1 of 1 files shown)


benhoyt:
@stephanlachnit Can you please review these inih meson.build changes?

stephanlachnit:
LGTM

PR #159, opened 2023-10-13
Use fopen_s() with MSVC to silence its security warnings

HappySeaFox:
```
warning C4996: 'fopen': This function or variable may be
 unsafe. Consider using fopen_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online
 help for details.
```

### DIFF
--- a/ini.c
+++ b/ini.c
@@ -258,9 +258,15 @@ int ini_parse(const char* filename, ini_handler handler, void* user)
     FILE* file;
     int error;
 
+#ifdef _MSC_VER
+    errno_t err = fopen_s(&file, filename, "r");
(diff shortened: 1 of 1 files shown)


benhoyt:
Thanks for the suggestion. However, I don't think there's any real problem with `fopen`, and it's much more portable. I'd rather not litter (even more :-) `#ifdef`s around the code. So I think MSVC users with its over-aggressive warning should something [like this](https://stackoverflow.com/a/14387/68707) instead:

```
#define _CRT_SECURE_NO_DEPRECATE
#include "ini.h"
```

PR #165, opened 2024-03-26
INIReader: change variable visibility from private to protected

xkzl:
@benhoyt I am proposing to change variable visibility allowing for full feature inheritance.

### DIFF
--- a/cpp/INIReader.h
+++ b/cpp/INIReader.h
@@ -96,7 +96,7 @@ public:
     // Return true if a value exists with the given section and field names.
     INI_API bool HasValue(const std::string& section, const std::string& name) const;
 
-private:
+protected:
     int _error;
     std::map<std::string, std::string> _values;
     static std::string MakeKey(const std::string& section, const std::string& name);


benhoyt:
Hi @xkzl I'm not opposed to this, though it's been a while since I've done C++. Are there any other implications other than making these fields accessible to subclasses? Can you please provide a small example of how this would be used?

xkzl:
Hi @benhoyt, sorry for my late reply. You are right, there is no other implications than making fields and methods accessible to subclasses. Unless, the variable content is very specific (void* variable manipulation that only the mother class knows about for instance), it is okay to turn this as protected.

As an example, I would mention a custom INIReader I have been designing to retain the section and fields information.
Her eis the header file:

```
// Read an INI file into easy-to-access name/value pairs.

// inih and INIReader are released under the New BSD license (see LICENSE.txt).
// Go to the project home page for more info:
//
// http://code.google.com/p/inih/

#ifndef __INIREADER_H__
#define __INIREADER_H__

#include <map>
#include <set>
#include <string>

// Read an INI file into easy-to-access name/value pairs. (Note that I've gone
// for simplicity here rather than speed, but it should be pretty decent.)
class INIReaderPlus: public INIReader 
{
public:
    // Returns all the section names from the INI file, in alphabetical order, but in the
    // original casing
    std::set<std::string> GetSections() const;

    // Returns all the field names from a section in the INI file, in alphabetical order,
    // but in the original casing. Returns an empty set if the field name is unknown
    std::set<std::string> GetFields(std::string section) const;

protected:
    int _error;
    std::map<std::string, std::string> _values;

    std::set<std::string> _sections;
    std::map<std::string, std::set<std::string>*> _fields;
    static std::string MakeKey(std::string section, std::string name);
    static int ValueHandler(void* user, const char* section, const char* name,
                            const char* value);
};

#endif  // __INIREADER_H__
```

I can take advantage of your modifications and also customize the class, without protected elements I would have to rewrite the whole implementation (that I haven't studied). Doing so it helps factorizing.

benhoyt:
@mayak-dev @jcormier @vrresto @evorw You folks are the last few people who've touched INIReader (the C++ class) -- any objections to making these internal fields `protected` rather than `private`? If not I'll go ahead and merge this later this week.

mayak-dev:
Sounds fine to me!

jjcf89:
Seems reasonable

benhoyt:
Thanks for your replies, @mayak-dev and @jjcf89. Merging.

PR #170, opened 2024-05-07
Add meson test driver

matoro:
Switches tests to use "meson test" instead of a shell script.  This also means that they now fully respect user-specified toolchain (e.g. clang compiler, custom CFLAGS, etc).  Running C++ test is conditional on enabling INIReader support in the build.

```
 1/16 test_multi                               OK              0.05s
 2/16 test_multi_max_line                      OK              0.04s
 3/16 test_single                              OK              0.04s
 4/16 test_disallow_inline_comments            OK              0.03s
 5/16 test_stop_on_first_error                 OK              0.03s
 6/16 test_handler_lineno                      OK              0.02s
 7/16 test_heap                                OK              0.06s
 8/16 test_string                              OK              0.06s
 9/16 test_heap_max_line                       OK              0.05s
10/16 test_heap_realloc                        OK              0.05s
11/16 test_heap_realloc_max_line               OK              0.05s
12/16 test_heap_string                         OK              0.04s
13/16 test_call_handler_on_new_section         OK              0.04s
14/16 test_allow_no_value                      OK              0.03s
15/16 test_alloc                               OK              0.02s
16/16 test_INIReaderExample                    OK              0.02s

Ok:                 16
Expected Fail:      0
Fail:               0
Unexpected Pass:    0
Skipped:            0
Timeout:            0
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
 
     - name: Run Diff Tests
       run: |
@@ -20,3 +20,14 @@ jobs:
         cd ../examples
         ./cpptest.sh
         git diff --exit-code
+
+  build-meson:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
+    - uses: BSFishy/meson-build@v1.0.3
+      with:
+        action: test
+        meson-version: 1.4.1
--- a/examples/meson.build
+++ b/examples/meson.build
@@ -1,0 +1,10 @@
+runtest = files(join_paths(meson.project_source_root(), 'tests', 'runtest.sh'))
+
+tests = {
+  'INIReaderExample': { 'args': [] },
+}
+
+foreach name, properties : tests
+  exe = executable('unittest_' + name, src_inih, src_INIReader, 'INIReaderExample.cpp', cpp_args : ['-Wall', properties['args']])
+  test('test_' + name, runtest, depends : [exe], args : [files('cpptest.txt'), exe.full_path()])
+endforeach
--- a/meson.build
+++ b/meson.build
@@ -2,7 +2,8 @@ project('inih',
     ['c'],
     license : 'BSD-3-Clause',
     version : '58',
-    default_options : ['cpp_std=c++11']
+    default_options : ['cpp_std=c++11'],
+    meson_version: '>=0.56.0'
 )
 
 #### options ####
@@ -71,8 +72,10 @@ endif
 #### inih ####
 inc_inih = include_directories('.')
 
+src_inih = files('ini.c')
+
 lib_inih = library('inih',
-    ['ini.c'],
+    [src_inih],
     include_directories : inc_inih,
     c_args : [arg_static, extra_args],
     install : distro_install,
@@ -96,13 +99,17 @@ inih_dep = declare_dependency(
     include_directories : inc_inih
 )
 
+subdir('tests')
+
 #### INIReader ####
 if get_option('with_INIReader')
     add_languages('cpp')
     inc_INIReader = include_directories('cpp')
 
+    src_INIReader = files(join_paths('cpp', 'INIReader.cpp'))
+
     lib_INIReader = library('INIReader',
-        ['cpp/INIReader.cpp'],
+        src_INIReader,
         cpp_args : extra_args,
         include_directories : inc_INIReader,
         dependencies : inih_dep,
@@ -126,4 +133,6 @@ if get_option('with_INIReader')
         include_directories : inc_INIReader,
         compile_args : extra_args
     )
+
+    subdir('examples')
 endif
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -1,0 +1,25 @@
+runtest = files(join_paths(meson.project_source_root(), 'tests', 'runtest.sh'))
+
+tests = {
+  'multi': { 'args': [] },
+  'multi_max_line': { 'args': ['-DINI_MAX_LINE=20'] },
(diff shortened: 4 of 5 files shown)


benhoyt:
I appreciate your efforts here. However, I'd rather not *require* meson as part of the test process of inih, partly because it's a tool I'm not at all familiar with so can't effectively review, and partly because I don't want me or others working on inih to have to install a separate tool (meson) to run the tests.

matoro:
> I appreciate your efforts here. However, I'd rather not _require_ meson as part of the test process of inih, partly because it's a tool I'm not at all familiar with so can't effectively review, and partly because I don't want me or others working on inih to have to install a separate tool (meson) to run the tests.

Thank you for looking; would you be interested in accepting if I restored the original shell scripts so that both methods can coexist?  Of course they could get out of sync, I understand if you'd rather not due to that concern.

benhoyt:
Yes, actually, I'd be willing for them to coexist - good idea.

benhoyt:
I also realised we should run the CI workflow on PR branches - feel free to make that change too.

benhoyt:
Oh wait, I guess we do already.

.github/workflows/tests.yml
```diff
@@ -20,3 +20,12 @@ jobs:
         cd ../examples
         ./cpptest.sh
         git diff --exit-code
+  build-meson:
```
benhoyt, line 23: Nit: can you please add a blank line above `build-meson:`?

.github/workflows/tests.yml
```diff
@@ -20,3 +20,12 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
```
benhoyt, line 27: While we're adding this, could we use the latest, `@v4`?

.github/workflows/tests.yml
```diff
@@ -20,3 +20,12 @@ jobs:
...
+    - uses: actions/setup-python@main
+    - uses: BSFishy/meson-build@master
```
benhoyt, line 29: I'd prefer to pin to a specific version, looking at them briefly, I think `setup-python@v5` and `BSFishy/meson-builder@v1.0.3`.

runtest.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
benhoyt, line 1: Can we avoid having this script, or at least, avoid putting it in the root directory? It looks like it basically just runs a `diff` command? Some options:

1) Include it directly in the meson.build files (I'm not sure what's possible with meson).
2) Put it at `tests/runtest.sh`, and reference it from `examples/meson.build`.
3) Copy it both places (it's tiny).

runtest.sh
```diff
@@ -0,0 +1,6 @@
...
+
+set -euo pipefail
+
+cd "$(dirname "${1}")"
+diff "${1}" <("${2}")
```
benhoyt, line 6: How does this actually work? I guess it's running something like `diff baseline_multi.txt`, but what's the `<(...)` redirection for?

tests/meson.build
```diff
@@ -0,0 +1,46 @@
+runtest = files(join_paths(meson.project_source_root(), 'runtest.sh'))
+
+unittest_multi = executable('unittest_multi', src_inih, 'unittest.c', c_args : ['-Wall'])
+test('test_multi', runtest, depends : [unittest_multi], args : [files('baseline_multi.txt'), unittest_multi.full_path()])
+
+unittest_multi_max_line = executable('unittest_multi_max_line', src_inih, 'unittest.c', c_args : ['-Wall', '-DINI_MAX_LINE=20'])
+test('test_multi_max_line', runtest, depends : [unittest_multi_max_line], args : [files('baseline_multi_max_line.txt'), unittest_multi_max_line.full_path()])
```
benhoyt, line 7: I don't know what's possible with meson, but it looks like there's data structures. Is it possible to write these as something like a table-driven test, and only include the variables (things that are different) in the table data. Maybe something like:

```
tests = [
    ['multi', ['-Wall']],
    ['multi_max_line', ['-Wall', '-DINI_MAX_LINE=20']],
    ['single', ['-Wall', '-DINI_ALLOW_MULTILINE=0']],
    ...
]
for name, c_args in tests:
    test('test_'+name, runtest, ...)
```

runtest.sh
matoro: Yup, that's executing the command in the second argument, piping its stdout to a named pipe, and then comparing those contents against the baseline file.  I got this trick from [this answer](https://unix.stackexchange.com/a/155807) (but only one of the inputs needs to be a command instead of both)

.github/workflows/tests.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
```
benhoyt, line 14: Thanks!

examples/meson.build
```diff
@@ -0,0 +1,11 @@
+runtest = files(join_paths(meson.project_source_root(), 'tests', 'runtest.sh'))
+
+tests = {
+  'INIReaderExample': { 'args': [] },
+}
+
+foreach name, properties : tests
+  test_src = 'src' in properties ? properties['src'] : 'INIReaderExample.cpp'
```
benhoyt, line 8: Given it's always `INIReaderExample.cpp` here, this was confusing. What about just removing this line and hard-coding `INIReaderExample.cpp` on the next line here?

meson.build
```diff
@@ -2,7 +2,8 @@ project('inih',
     ['c'],
     license : 'BSD-3-Clause',
     version : '58',
-    default_options : ['cpp_std=c++11']
+    default_options : ['cpp_std=c++11'],
+    meson_version: '>=0.56.0'
```
benhoyt, line 6: Just for my benefit, what change or option here requires >=0.56, if any?

benhoyt (approved):
Nice, this is much cleaner, thanks! One minor suggestion, take it or leave it.

PR #175, opened 2024-07-24
Update INIReader.cpp

xkzl:
Update INIReader::INIReader prototype implementation to match header implementation.
This update is removing the warning when generating Doxygen documentation.

### DIFF
--- a/cpp/INIReader.cpp
+++ b/cpp/INIReader.cpp
@@ -17,7 +17,7 @@
 
 using std::string;
 
-INIReader::INIReader(const string& filename)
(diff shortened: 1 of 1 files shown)


benhoyt:
Thanks for the contribution, but I don't think I'll merge this. If we were going to change it here, we should change it everywhere and remove the `using std::string` line (but I'd rather not do that -- the `using` declaration is intentional as it keeps the code shorter). If Doxygen can't figure out that `std::string` and `string` (with `using std::string`) are the same thing, that's a Doxygen problem.

PR #178, opened 2024-10-23
Add preprocessor exclusions when INI_ALLOW_MULTILINE=0

isidroas:
Exclude the `prev_name` variable and the ini_strncpy0() call.

This decreases the program size, the execution time and the stack usage when that flag is off.

But that's barely noticeable. It's more useful for progressive understanding.
When I first came across with this code, I removed (or collapsed) all the `ifdef` blocks to learn the simpler version first.


benhoyt (approved):
I don't love how many `#if`s there are now, but on balance, I'm okay with this change. Thanks!

PR #179, opened 2024-10-23
tests: define only once the compiler and flag

isidroas:
- **tests: define only once the compiler and flag**
    This way you could play with them easily. For example:
    - use clang instead gcc
    - -Wextra
    - -fsanitize=address

- **check CC environ and fallback to gcc**

### DIFF
--- a/tests/unittest.sh
+++ b/tests/unittest.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-CC="gcc -Wall"
+CC="${CC:-gcc} -Wall"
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
Works for me, thanks!

```
$ time CC=gcc ./unittest.sh

real	0m0.841s
user	0m0.574s
sys	0m0.257s
$ time CC=tcc ./unittest.sh

real	0m0.141s
user	0m0.070s
sys	0m0.069s
```

PR #180, opened 2024-10-25
Test that long sections are truncated

isidroas:
- **test that long sections are truncated**
- **generated .txt executing unittest.sh**
- **expected result**

Should we revert the commit 9f0a1d2199b1a9f0a20003d848918a881a2d91e4 to prevent the CI from failing?

It seems related with a more general issue ( #145 )

### DIFF
--- a/tests/baseline_allow_no_value.txt
+++ b/tests/baseline_allow_no_value.txt
@@ -69,3 +69,6 @@ duplicate_sections.ini: e=0 user=108
 ... [section1]
 ... key1=val1;
 no_value.ini: e=0 user=109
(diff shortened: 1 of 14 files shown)


benhoyt:
Yeah, I've reverted https://github.com/benhoyt/inih/commit/9f0a1d2199b1a9f0a20003d848918a881a2d91e4 as I think it's actually the correct behaviour, because `len(section) > len(INI_MAX_LINE)` for these tests. I'd like to improve the tests so that tests aren't so dependent on each other, but I'll do that later. Merging this now.

benhoyt:
Thanks for the contribution!

PR #181, opened 2024-10-29
Replicate and fix long lines 

isidroas:


isidroas:
Closing because I am not happy with my solution. However, the test could be reused.

PR #183, opened 2024-12-06
Use 'fopen_s' for Security Reasons

Tristan-BS:
**Description**
This pull request replaces the `fopen` function with the safer `fopen_s` function to enhance security and error handling. The `fopen_s` function provides better safeguards against potential vulnerabilities associated with file opening.

**Changes Made**

- Replaced `fopen` with `fopen_s` for opening files.

**Justification**
- [fopen_s on MSDN](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-s-wfopen-s?view=msvc-160)

### DIFF
--- a/ini.c
+++ b/ini.c
@@ -264,14 +264,13 @@ int ini_parse(const char* filename, ini_handler handler, void* user)
     FILE* file;
     int error;
 
-    if (fopen_s(&file, filename, "r") != 0) {
+    file = fopen(filename, "r");
+    if (!file)
         return -1;
-    }
     error = ini_parse_file(file, handler, user);
     fclose(file);
     return error;
 }
-
 
 /* An ini_reader function to read the next line from a string buffer. This
    is the fgets() equivalent used by ini_parse_string(). */
--- a/ini.c
+++ b/ini.c
@@ -264,9 +264,9 @@ int ini_parse(const char* filename, ini_handler handler, void* user)
     FILE* file;
     int error;
 
-    file = fopen(filename, "r");
-    if (!file)
+    if(fopen_s(&file, filename, "r") != 0) {
         return -1;
+    }
     error = ini_parse_file(file, handler, user);
     fclose(file);
     return error;
--- a/ini.c
+++ b/ini.c
@@ -264,13 +264,14 @@ int ini_parse(const char* filename, ini_handler handler, void* user)
     FILE* file;
     int error;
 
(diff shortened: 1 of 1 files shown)


Tristan-BS:
PS: I initially committed the change to use `fopen_s` for better security, but reverted it due to a potential issue. After verifying the fix, I recommitted the change. Please don't be alarmed by the revert.

benhoyt:
Can you tell me what security problem this is fixing? I've looked briefly into the difference between `fopen` and `fopen_s`, and if you check the return value of `fopen` (and the mode is "r"), there doesn't seem to be any advantage. In addition, `fopen_s` is less common and less portable (only standardised in C11).

Tristan-BS:
Look at this link:
- https://stackoverflow.com/questions/2575116/fopen-fopen-s-and-writing-to-files
-  https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-s-wfopen-s?view=msvc-170
- https://en.cppreference.com/w/%20%20c/io/fopen


The answer from ChatGPT for example:

The fopen_s function is a safer alternative to the standard fopen in C and C++ for the following reasons:

1. Enhanced Security: fopen_s performs additional checks, such as validating that pointers (e.g., filename, mode, stream pointer) are not NULL. This prevents common vulnerabilities like undefined behavior when invalid arguments are passed.


2. Explicit Error Handling: Unlike fopen, which returns NULL on failure, fopen_s provides an error code indicating the nature of the issue. Additionally, it ensures the output pointer is set to NULL if the file operation fails, making error detection more robust.


3. Better Standards Compliance: fopen_s is part of the optional C11 bounds-checking interface. It was designed to reduce risks of programming errors, such as buffer overflows, that can occur with fopen.


4. Reduced Risk of Undefined Behavior: By explicitly enforcing proper usage and input validation, fopen_s minimizes cases where incorrect usage could result in hard-to-debug issues.



While fopen_s improves security and reliability, it is not universally portable because it is part of Microsoft's secure runtime extensions and the C11 optional standard library.

Tristan-BS:
If portability across different compilers and platforms is required, use fopen and close this MR.

benhoyt:
Thanks. I don't actually think `fopen_s` will help security in this use case here in inih, and it's less portable, so I'd rather not switch to this. Closing.

Tristan-BS:
Alright, thanks for the inspection and investigate

PR #186, opened 2025-03-01
Add INIReader Sections and Keys methods

Ishan09811:
This PR adds a SetValue function to the INIReader class, allowing users to modify INI values.

### DIFF
--- a/cpp/INIReader.h
+++ b/cpp/INIReader.h
@@ -89,6 +89,8 @@ public:
     // and valid false values are "false", "no", "off", "0" (not case sensitive).
     INI_API bool GetBoolean(const std::string& section, const std::string& name, bool default_value) const;
 
+    INI_API void SetValue(const std::string& section, const std::string& name, const std::string& value);
+
     // Return true if the given section exists (section must contain at least
     // one name=value pair).
     INI_API bool HasSection(const std::string& section) const;
--- a/cpp/INIReader.h
+++ b/cpp/INIReader.h
@@ -91,9 +91,6 @@ public:
     // and valid false values are "false", "no", "off", "0" (not case sensitive).
     INI_API bool GetBoolean(const std::string& section, const std::string& name, bool default_value) const;
 
-    // Sets value of the specified key
-    INI_API void SetValue(const std::string& section, const std::string& name, const std::string& value);
-
     // Return a newly-allocated vector of all section names, in alphabetical order.
     INI_API std::vector<std::string> Sections() const;
 
--- a/cpp/INIReader.h
+++ b/cpp/INIReader.h
@@ -15,6 +15,7 @@
 #include <map>
 #include <string>
 #include <cstdint>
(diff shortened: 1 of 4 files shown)


cpp/INIReader.h
```diff
@@ -89,6 +89,15 @@ class INIReader
     // and valid false values are "false", "no", "off", "0" (not case sensitive).
     INI_API bool GetBoolean(const std::string& section, const std::string& name, bool default_value) const;
 
+    // Sets value of the specified key
+    INI_API void SetValue(const std::string& section, const std::string& name, const std::string& value);
+
+    // Get a list of section names
+    INI_API std::vector<std::string> Sections() const;
+
+    // Get a list of keys for a given section
```
benhoyt, line 98:
```suggestion
    // Return a newly-allocated vector of keys in the given section, in alphabetical order.
```

examples/INIReaderExample.cpp
```diff
@@ -24,5 +24,20 @@ int main()
...
+    std::vector<std::string> sections = reader.Sections();
+    for (std::vector<std::string>::const_iterator it = sections.begin(); it != sections.end(); ++it) {
+        std::cout << "- " << *it << "\n";
+    }
+    
+    for (std::vector<std::string>::const_iterator it = sections.begin(); it != sections.end(); ++it) {
+        std::cout << "Keys in section [" << *it << "]:\n";
+        std::vector<std::string> keys = reader.Keys(*it);
+        for (std::vector<std::string>::const_iterator kit = keys.begin(); kit != keys.end(); ++kit) {
+            std::cout << "  - " << *kit << "\n";
```
benhoyt, line 38: Let's use no indent to match the above:
```suggestion
            std::cout << "- " << *kit << "\n";
```

benhoyt:
Looks like tests are failing, too. See https://github.com/benhoyt/inih/actions/runs/13603098149/job/38031412682?pr=186

Ishan09811:
i actually needed the SetValue function and will be very useful for other users also so I will still request you to keep SetValue function but if you don't then no worry i will make a subclass for my usecase and I am happy to contribute

benhoyt:
> i actually needed the SetValue function and will be very useful for other users also so I will still request you to keep SetValue function but if you don't then no worry i will make a subclass for my usecase and I am happy to contribute

Yep, you'll need to subclass for that, sorry.

Ishan09811:
Ok thanks for your quick response and now the test hopefully will work fine.

Ishan09811:
do you know why is it failing because it is not outputing any error message so I can't find what actually causing the error

Ishan09811:
finally fixed the issue actually before I thought that it is failed when compiling but when actually I saw it carefully then i understand that it is trying find any difference between the printed message and cpptest.txt so now i fixed the issue lol

benhoyt:
Thanks for the contribution!

PR #192, opened 2025-04-03
meson: fix tests under Windows

neheb:


neheb:
@bgilbert nope. Probably the solution is to replace runtest.sh with python.

neheb:
hrm sh works. a couple of seconds slower. not too bad though.

neheb:
Interesting...

benhoyt (approved):

PR #196, opened 2025-06-09
add ini_parse_string_length() that avoids internal strlen(), and ease…

lihuiba:
add ini_parse_string_length() that avoids internal strlen(), and ease interacting with C++ std::string_view.

### DIFF
--- a/ini.c
+++ b/ini.c
@@ -316,6 +316,7 @@ int ini_parse_string(const char* string, ini_handler handler, void* user) {
     return ini_parse_string_length(string, strlen(string), handler, user);
 }
 
+/* See documentation in header file. */
 int ini_parse_string_length(const char* string, size_t length,
                             ini_handler handler, void* user) {
     ini_parse_string_ctx ctx;
--- a/ini.c
+++ b/ini.c
@@ -313,10 +313,15 @@ static char* ini_reader_string(char* str, int num, void* stream) {
 
 /* See documentation in header file. */
 int ini_parse_string(const char* string, ini_handler handler, void* user) {
+    return ini_parse_string_length(string, strlen(string), handler, user);
+}
+
+int ini_parse_string_length(const char* string, size_t length,
+                            ini_handler handler, void* user) {
     ini_parse_string_ctx ctx;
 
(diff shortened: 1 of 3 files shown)


ini.c
```diff
@@ -313,10 +313,15 @@ static char* ini_reader_string(char* str, int num, void* stream) {
 
 /* See documentation in header file. */
 int ini_parse_string(const char* string, ini_handler handler, void* user) {
+    return ini_parse_string_length(string, strlen(string), handler, user);
+}
+
+int ini_parse_string_length(const char* string, size_t length,
```
benhoyt, line 319:
```suggestion
/* See documentation in header file. */
int ini_parse_string_length(const char* string, size_t length,
```

ini.h
```diff
@@ -92,10 +92,15 @@ INI_API int ini_parse_stream(ini_reader reader, void* stream, ini_handler handle
                      void* user);
 
 /* Same as ini_parse(), but takes a zero-terminated string with the INI data
-instead of a file. Useful for parsing INI data from a network socket or
-already in memory. */
+   instead of a file. Useful for parsing INI data from a network socket or
+   already in memory. */
```
benhoyt, line 96:
```suggestion
   which is already in memory. */
```

ini.h
```diff
@@ -92,10 +92,15 @@ INI_API int ini_parse_stream(ini_reader reader, void* stream, ini_handler handle
...
 INI_API int ini_parse_string(const char* string, ini_handler handler, void* user);
 
+/* Same as ini_parse_string(), but takes a string and its length, avoiding
+   strlen(). Useful for parsing INI data from a network socket or already
+   in memory, or interfacing with C++ std::string_view. */
```
benhoyt, line 101:
```suggestion
/* Same as ini_parse_string(), but takes a string and its length, avoiding
   strlen(). Useful for parsing INI data from a network socket or which is
   already in memory, or interfacing with C++ std::string_view. */
```

benhoyt (approved):
Thanks for this, nice addition in keeping with the spirit of the library. Couple of nit comments which I'll apply.

PR #197, opened 2025-07-09
Fix bug where name only options are not processed after an error has occurred

msquire:
After an error has occurred the expected behaviour is for parsing to continue unless the `INI_STOP_ON_FIRST_ERROR` option is set. However when `INI_ALLOW_NO_VALUE` is set name only options are currently silently ignored after an error. 

This fixes that minor bug and adds a new test case because it is not exposed by any of the existing ones.


benhoyt (approved):
Thanks for the fix! And the new test file looks reasonable to me.

PR #198, opened 2025-07-09
Optimise ini_rstrip()

msquire:
The `ini_rstrip()` function finds the end of the string using `strlen()`, which is fine. However, everywhere the function is called the code has already discovered a pointer to the end of the string by other means.

This optimises out the `strlen()` call by passing the pointer to the end of the string instead. All the tests pass as expected.

If you are minded to merge this pull request you might also consider inlining the trivial `ini_rstrip()` and `ini_lskip()` functions. I created [this branch](https://github.com/msquire/inih/tree/optimise-rstrip-inline) (branched of the one in this PR) which inlines these functions if the compiler supports inlining (i.e. it is C99 or later).  You might also consider inlining the `ini_find_chars_or_comment()` and `ini_strncpy0()` functions although they are slightly less trivial so I didn't include them in the branch.

If you would like to include the inlining, let me know and I will merge it into the branch in this PR.


benhoyt:
Thanks for the contribution. I'll try to take a look in the next few days!

msquire:
No worries! Let me know your thoughts and if you want any merges/changes after you get a chance to review it.

benhoyt:
This looks good to me, thanks!

I'd prefer not to merge the `inline` thing. Wouldn't a good compiler automatically inline small `static` functions anyway, given the right compiler optimisation options? In any case, inih is intended to be (mostly) optimised for size.

I'll fix the `//` comment in a follow-up commit.

PR #210, opened 2026-01-30
Assert on NULL arguments

### LINKED ISSUE #209: Crash in ini_parse_string() due to NULL pointer dereference
I found a NULL pointer dereference in ini_parse_string() while fuzzing libinih
using OSS-Fuzz (libFuzzer + AddressSanitizer).

The crash occurs when ini_parse_string() is called with certain malformed input
and results in a segmentation fault due to a NULL read.

Environment:
- Fuzzing engine: libFuzzer
- Sanitizer: AddressSanitizer (ASan)
- OS: Linux (OSS-Fuzz environment)

Crash summary:
AddressSanitizer: SEGV (null pointer dereference)
Function: ini_parse_string_length()
ini.c

Stack trace:
ini_parse_string_length
ini_parse_string
LLVMFuzzerTestOneInput

Reproducer:
The crash is reproducible using the following input (hex):
3f 74 3d

Base64:
P3Q9

This issue was discovered during fuzz testing and may be exploitable depending
on usage context.

DimitriPapadopoulos:
Fixes #209.


ini.c
```diff
@@ -123,6 +124,8 @@ int ini_parse_stream(ini_reader reader, void* stream, ini_handler handler,
     char abyss[16];  /* Used to consume input when a line is too long. */
     size_t abyss_len;
 
+    assert(reader != NULL && stream != NULL && handler != NULL);
```
benhoyt: Thanks. I'd prefer these as separate asserts on separate lines, so that it's easy to see in a stack trace / diagnostic output which variable is NULL from the line number.

benhoyt (changes requested):

benhoyt (approved):
Thanks for this.

PR #212, opened 2026-07-14
Make section and name buffer limits configurable

94xhn:
`ini_parse_stream()` stores the current section and previous key name in fixed-size buffers. The existing 50-byte name buffer means a 50-character key is truncated on a multiline continuation callback.

This keeps the existing defaults and stack usage unchanged while making both limits configurable:

- Add public `INI_MAX_SECTION` and `INI_MAX_NAME` macros in `ini.h`, both defaulting to 50 bytes including the NUL terminator.
- Use those macros for the parser's saved section and name buffers.
- Let applications that accept longer names opt in at compile time without tying these buffers to `INI_MAX_LINE`.

Tests cover both configurations:

- The default build asserts both limits remain 50 and preserves a 49-character multiline key.
- A build with `INI_MAX_SECTION=51` and `INI_MAX_NAME=51` preserves 50-character section and key names across both callbacks.
- Both focused targets compile with GCC 8.1 using `-std=c99 -Wall -Wextra -Werror` and pass.
- The existing default `unittest.c` output remains byte-for-byte identical to `baseline_multi.txt`.


benhoyt:
Hi @94xhn, thanks for this submission. I'd rather not take this in its current form. I'd consider a change that added INI_MAX_SECTION and INI_MAX_NAME to ini.h without changing the defaults of 50 and 50. However, 50 chars is a rather long key name -- what types of files or keys are you hitting that limit with? Can you provide some samples?

94xhn:
Thanks for the direction. I updated the PR in f12138b to add INI_MAX_SECTION and INI_MAX_NAME in ini.h, both defaulting to 50, so the default stack usage is unchanged. The focused tests now cover the unchanged 49-character default boundary and an explicit 51-byte configuration that preserves 50-character section and key names across multiline callbacks. I also verified the existing default unittest output remains byte-for-byte identical to baseline_multi.txt. I do not have a production configuration sample with a key this long; I found this by boundary-testing the multiline callback behavior, where the first and continuation callbacks reported different names for the same parsed key. Making the limit opt-in is a better fit than raising the default.

benhoyt:
I've decided I'd prefer not to make these configurable or add new #define's at this point. But thanks for the submission.

